PR #8, opened 2022-08-08
feat: render `react`

zenorocha:
the whole point is having the SDK handle that part, instead of introducing another dependency that the user needs to know

### DIFF
--- a/src/klotty.ts
+++ b/src/klotty.ts
@@ -1,6 +1,6 @@
 import axios, { AxiosInstance, AxiosResponse } from 'axios';
 import { SendEmailData } from './interfaces';
-import reactEmail from 'react-email';
+import { render } from 'react-email';
 
 const VERSION = '0.2.1';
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,7 +27,8 @@
   },
   "homepage": "https://github.com/klotty/klotty-node#readme",
   "dependencies": {
-    "axios": "0.27.2"
+    "axios": "0.27.2",
+    "react-email": "^1.0.2"
```
bukinoshita: need to pin the dependency version (`yarn add -E react-email`)

src/klotty.ts
```diff
@@ -35,6 +36,11 @@ export class Klotty {
   async sendEmail(data: SendEmailData): Promise<AxiosResponse> {
     try {
       const path = `${this.baseUrl}/email`;
+
+      if (data.data.react) {
+        data.data.html = reactEmail.render(data.data.react);
```
bukinoshita: It feels weird to have `data.data`, do you know why?

PR #35, opened 2022-12-27
chore: Update package version to 0.9.0

vinicoder:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "resend",
-  "version": "0.8.0",
+  "version": "0.8.1",
```
zenorocha: Should be 0.9.0 since it contains a new feature

PR #83, opened 2023-04-08
feat: Map error response type

ivoilic:
Rather than just returning string/number these new types give developers a better idea of what errors the might receive. 

This setup was inspired by how TRPC types it's codes: https://github.com/trpc/trpc/blob/main/packages/server/src/rpc/codes.ts

I think it would be great and take this one step further by unifying the return types. There's no reason type, status, and message couldn't exist on the returned success value too.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -54,7 +54,7 @@ export interface SendEmailRequest {
 }
 
 export const RESEND_ERROR_CODES_BY_KEY = {
-  missing_required_fields: 400,
+  missing_required_fields: 422,
   missing_api_key: 401,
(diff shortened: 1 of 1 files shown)


src/interfaces.ts
```diff
@@ -53,13 +53,28 @@ export interface SendEmailRequest {
   tags?: Tag[];
 }
 
+export const RESEND_ERROR_CODES_BY_KEY = {
+  missing_required_fields: 400,
```
bukinoshita:
```suggestion
  missing_required_fields: 422,
```

bukinoshita:
Thanks a lot @ivoilic ❤️

PR #131, opened 2023-07-08
feat: Add headers parameter to CreateEmailOptions type

jayanratna:
> **Note**
> This PR depends on #130 to function correctly.

This PR adds the new `headers` parameter to the CreateEmailOptions type.

### DIFF
--- a/src/resend.spec.ts
+++ b/src/resend.spec.ts
@@ -1,6 +1,7 @@
 import { Resend } from './resend';
 import MockAdapater from 'axios-mock-adapter';
 import axios from 'axios';
+import { CreateEmailOptions } from './emails/interfaces';
 
 const mock = new MockAdapater(axios);
 const resend = new Resend('re_924b3rjh2387fbewf823');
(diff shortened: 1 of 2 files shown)


kyldvs:
Thanks! I'm currently blocked on using resend due to the inability to add headers.

src/resend.spec.ts
```diff
@@ -160,4 +167,38 @@ describe('Resend', () => {
...
+      from: 'admin@resend.com',
+      to: 'bu@resend.com',
+      headers: {
+        'X-Entity-Ref-ID': '123',
+      },
+      created_at: '123',
+    });
+
+    const data = await resend.emails.send(payload);
+    expect(data).toMatchInlineSnapshot(`
```
bukinoshita, line 186: CreateEmailResponse, only returns `id`

jayanratna: Thanks for catching that. I'll get that updated ASAP.

bukinoshita (approved):

PR #146, opened 2023-07-27
feat: Add missing fields to the `GetEmailResponse` type

abubuwe:
1. As seen in the [API documentation](https://resend.com/docs/api-reference/emails/retrieve-email), the `GetEmailResponse` type is missing some fields that are useful to access. (Side note: the API docs appear to be wrong in that the `bcc, cc, reply_to` fields do return `null` and not the advertised `[null]`).
This PR adds those missing fields.

2. It also seems that the API returns fields with `null`, rather than omitting them or with the value `undefined`, so this PR also updates the types to match this behaviour. 

3. A small addition fix is that the GetEmailResponse type is now exported in the `index.ts`. 

Please feel free to point out any issues or things that I may have failed to consider in this PR.


bukinoshita:
can you run `yarn format`?

package.json
```diff
@@ -16,7 +16,7 @@
     "test:watch": "jest --watch",
     "lint": "eslint --fix --ext .ts ./src",
     "format:check": "prettier --check \"**/*.{ts,tsx}\"",
-    "format": "prettier --write .",
+    "format": "prettier --write \"**/*.{ts,tsx}\"",
```
bukinoshita: Can we remove this?

abubuwe: Removed 👍

bukinoshita (approved):

PR #149, opened 2023-08-04
refactor: Replace `axios` with `node-fetch`

bukinoshita:


src/resend.ts
```diff
@@ -30,113 +29,91 @@ export class Resend {
...
-        if (error?.response?.data) {
-          return Promise.reject(error.response.data);
-        }
-        return Promise.reject(error);
-      },
-    );
+    return await response.json();
   }
 
   async post<T>(path: string, entity?: any, options?: PostOptions): Promise<T> {
```
leonardomerlin: What do you think about initializing `options` with empty object (as the same approach as done for `GetOptions` and `PutOptions`)?

```suggestion
  async post<T>(path: string, entity?: any, options: PostOptions = {}): Promise<T> {
```

PR #224, opened 2023-09-30
feat: Add error handling and rejections

Mythie:
The prior version of the SDK would silently swallow errors that occurred returning the JSON response without any indication that a non-OK status code had been encountered.

We have now updated the SDK to instead throw when error responses are received with a custom error class being used for type narrowing.

Additionally we have updated the tests and types to be more correct utilising both type guards and the `satisfies` keyword.

> **🚨 This is a breaking change as we are now throwing errors, this may break the SDK for upstream users who weren't aware of the silent error swallowing.**


src/interfaces.ts
```diff
@@ -1,11 +1,16 @@
 export const RESEND_ERROR_CODES_BY_KEY = {
-  missing_required_fields: 422,
+  missing_required_field: 422,
+  invalid_access: 422,
+  invalid_parameter: 422,
```
vcapretz, line 4:
```suggestion
```
we don't actually return these errors from the API, we can remove them

vcapretz: btw, here's an updated list with the errors we do throw, we can update them in the SDK as well:
```
| 'missing_required_parameter'
    | 'missing_required_field'
    | 'invalid_attachment'
    | 'missing_api_key'
    | 'restricted_api_key'
    | 'not_found'
    | 'rate_limit_exceeded'
    | 'daily_quota_exceeded'
    | 'application_error'
    | 'validation_error'
    | 'security_error';
```

I'll update our Docs to make sure they are aligned in there as well 👍

src/resend.ts
```diff
@@ -40,7 +42,12 @@ export class Resend {
 
     if (!response.ok) {
       const error = await response.json();
-      return error;
+
+      if (isResendErrorResponse(error)) {
+        throw new ResendError(error.error.message, error.error.name, error.error.statusCode);
```
vcapretz: I would rather not throw errors instead, I believe we should return either `data` or `error` and have typescript set to something like this:
```
type ResponseWithError = { data: undefined; error: ErrorObject }
type SuccessfulResponse = { data: Data, error: undefined }
type Response = SuccessfulResponse | ResponseWithError
```

this way developers can correctly narrow the types when checking for success or errors 👍 


wdyt?

src/interfaces.ts
Mythie: I'm fairly certain I received both in my resend logs during setup. Relates to an invalid from address and a non validated domain 🤔

src/resend.ts
Mythie: I'm all for result types since they're more obvious. If doing so we should also export a type guard to help with inference 😃

vcapretz: great idea! is that something you can work on?

Mythie: Of course, I'll update this tomorrow morning 😄

src/interfaces.ts
Mythie: Fwiw invalid params definitely does appear in my resend logs from using a bad from address. Not sure if you can verify but I can certainly confirm it in my dashboard.

src/resend.ts
vcapretz: great! I really appreciate it, we are preparing a major release for this package and having better error handling is going to be essential, thanks for looking into it 😄

src/interfaces.ts
Mythie:
```
{
    "statusCode": 422,
    "name": "invalid_parameter",
    "message": "Invalid `from` field. The email address needs to follow the `email@example.com` or `Name <email@example.com>` format"
}
```

src/resend.ts
Mythie: G'day sorry I had updated but said change is quite breaking since you're changing actual response types.

By having:

```
{
   data: ...
   error: undefined
}
```

We are changing most everything in the SDK, I'm happy to proceed with these updates but it might also be a nice idea to stop and consider if we can do:

```
type Response = Success | Error
```

Which is more achievable without introducing breaking changes.

src/interfaces.ts
vcapretz: you're right! we do return this one 👍

src/resend.ts
vcapretz: hey @Mythie, yeah, let's start with something more simple for now, I'm happy with either so doing `Success | Error` is a good step! I would just be wary of actually throwing errors, so as long we are returning them instead it's all good

Mythie: Sorry for the delay! Made said updates 😄

Mythie:
Wooo 🙌🏻

PR #259, opened 2023-11-03
fix (build): problems with next builds that used the sdk

gabrielmfern:
## What does this change?

This just substitutes the use of `tsc` for build, and uses tsup instead.
This also adds specifications for esm and cjs interoperability on the `package.json` so
that imports/requires don't get mixed up.

## Why is it necessary?

The missing interoperability between esm and cjs caused problems on next builds
and due to that some issues were openned on `react-email` reporting problems on build.

For the record, here is the error that was happening before:

```
./node_modules/@react-email/components/node_modules/html-to-text/lib/html-to-text.mjs + 24 modules
Cannot get final name for export 'encodeXML' of ./node_modules/entities/lib/esm/index.js
```

Also, [here](https://github.com/resendlabs/react-email/issues/1015) is the original issue from react-email.

## How does this solve the issue?

After investigating for a bit and testing out a few things I found out the problem
was on the Resend SDK, and after searching found out about [this](https://github.com/vercel/next.js/issues/39375) issue on NextJS that said this kind of error was due to missing interoperability between esm and cjs.

Besides this evidence, I have tested and confirm this works by copying the dist and substituting the 
resend module with the newly compiled one and the build worked on an existing project.


gabrielmfern:
Fixes #256

PR #267, opened 2023-11-15
feat: Add audiences

bukinoshita:


Jonovono:
@bukinoshita is there a plan to make a release with this in it?

bukinoshita:
> @bukinoshita is there a plan to make a release with this in it?
> 
> edit: ah, I can just use the canary :)

Yup, you can use the canary version. Please let me know if you have any feedbacks on the API

PR #269, opened 2023-11-16
feat: api for custom headers and fetch implementation 

shahradelahi:
This PR introduces a new feature that allows users to modify fetch headers or use a custom fetch implementation. It uses [nodejs/undici](https://github.com/nodejs/undici) for making requests. According to the Node.js development team, the Undici module has better performance and compatibility.

#### Use an HTTP Proxy

Using a proxy is useful for those who are behind a corporate firewall and trying to reach out to the `resend` API server. Here is an implementation for custom `fetch` for using HTTPS proxy:

```typescript
import { Resend } from 'resend';
import { fetch, ProxyAgent } from 'undici';
import 'dotenv/config';

if (!process.env.RESEND_API_KEY) {
  throw new Error('RESEND_API_KEY env is not defined');
}

const resend = new Resend(process.env.RESEND_API_KEY, {
  fetch: (input, init) => {
    return fetch(input, {
      ...init,
      dispatcher: new ProxyAgent('http://localhost:8080')
    });
  }
});

export default resend;
```

Please review this pull request and let me know if you have any feedback. Thank you!


gabrielmfern:
Like this idea very much. Changed it so that the native `fetch` is the default one instead of undici's.

Also, seems like the types between the undici and fetch are not interchangeable, I don't think this can work very well without some tricky changes that we don't have the time for right now. Closing it for now, since it's been such a long time (very sorry for taking this long to review), but if you make them both interchangeable let me know, I'll reopen and review.

Thank you!

PR #291, opened 2024-01-15
feat: Delete contact by email

bukinoshita:


vcapretz (approved):

polidario:
Is this already in production? I'm using the remove() but the payload is still using id and audience id. Email is still not existing

image

PR #296, opened 2024-01-16
fix: Types

bukinoshita:


src/contacts/interfaces/create-contact-options.interface.ts
```diff
@@ -3,11 +3,11 @@ import { ErrorResponse } from '../../interfaces';
...
 
 export interface CreateContactOptions {
-  audience_id: string;
+  audienceId: string;
   email: string;
   unsubscribed?: boolean;
-  first_name?: string;
-  last_name?: string;
+  firstName?: string;
+  lastName?: string;
```
bukinoshita, line 10: These are breaking changes

src/contacts/contacts.ts
```diff
@@ -26,48 +34,46 @@ export class Contacts {
...
     return data;
   }
 
   async update(payload: UpdateContactOptions): Promise<UpdateContactResponse> {
     const data = await this.resend.patch<UpdateContactResponseSuccess>(
-      `/audiences/${payload.audience_id}/contacts/${payload.id}`,
-      payload,
+      `/audiences/${payload.audienceId}/contacts/${payload.id}`,
+      {
+        ...payload,
```
vcapretz, line 66: should we destruct `audienceId` in here? I don't think it's used in the API's body, but the way we have now will send extra keys in the body.

an even better option would be to explicitly send what we want (since we only care about 3 keys) 
```suggestion
        unsubscribed: payload.unsubscribed,
```

Issue #122, opened 2023-07-04
Export all types used as function params on index


QuentinFrc:
Could be developer friendly to export type directly form index to able to do this import

`import { Resend, CreateEmailOptions } from 'resend';`

rather than this one

`import { CreateEmailOptions } from 'resend/build/src/emails/interfaces';`

My use case of needing thoses types is not really interesting but I think other peoples would make a great things with that

ghosh:
This no longer works in `2.1.0`, seems like the package build was changed in #259

iBobik:
Just wanted to use the type, but cannot
<img width="833" alt="Bez názvu" src="https://github.com/resend/resend-node/assets/614232/7ea9b03c-e03d-449f-aba8-b313ec7fe290">

PR #305, opened 2024-01-19
add export for CreateEmailOptions

JClackett:


MartinXPN:
Would it be possible to also expose `CreateEmailResponse`?

nickfrosty:
Related: https://github.com/resend/resend-node/issues/122

fnielson1:
Exporting all the types would be great!

vcapretz:
thank you @JClackett for the PR! I opened a new one #396 to work on it, it's going to make sense to end up exporting all request and response interfaces, but we might iterate on exactly how we do it 🙏

PR #408, opened 2024-08-05
feat: Add schedule emails

felipevolpone:
This also introduces the ability to cancel a scheduled email

### DIFF
--- a/src/emails/emails.ts
+++ b/src/emails/emails.ts
@@ -8,6 +8,8 @@ import type {
   CreateEmailResponseSuccess,
 } from './interfaces/create-email-options.interface';
 import type {
+  CancelScheduleResponse,
+  CancelScheduleResponseSuccess,
   GetEmailResponse,
   GetEmailResponseSuccess,
 } from './interfaces/get-email-options.interface';
@@ -44,6 +46,13 @@ export class Emails {
   async get(id: string): Promise<GetEmailResponse> {
     const data = await this.resend.get<GetEmailResponseSuccess>(
       `/emails/${id}`,
+    );
+    return data;
+  }
+
+  async cancelSchedule(id: string): Promise<CancelScheduleResponse> {
+    const data = await this.resend.delete<CancelScheduleResponseSuccess>(
+      `/email/${id}/schedule`,
     );
     return data;
   }
--- a/src/emails/emails.ts
+++ b/src/emails/emails.ts
@@ -54,7 +54,7 @@ export class Emails {
 
   async cancel(id: string): Promise<CancelEmailResponse> {
     const data = await this.resend.post<CancelEmailResponseSuccess>(
-      `/email/${id}/cancel`,
+      `/emails/${id}/cancel`,
     );
     return data;
   }
--- a/src/emails/emails.ts
+++ b/src/emails/emails.ts
@@ -2,14 +2,16 @@ import { renderAsync } from '@react-email/render';
 import type * as React from 'react';
 import type { Resend } from '../resend';
 import type {
+  CancelEmailResponse,
+  CancelEmailResponseSuccess,
+} from './interfaces/cancel-email-options.interface';
+import type {
   CreateEmailOptions,
   CreateEmailRequestOptions,
   CreateEmailResponse,
   CreateEmailResponseSuccess,
 } from './interfaces/create-email-options.interface';
 import type {
-  CancelScheduleResponse,
-  CancelScheduleResponseSuccess,
   GetEmailResponse,
   GetEmailResponseSuccess,
 } from './interfaces/get-email-options.interface';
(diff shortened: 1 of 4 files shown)


src/emails/interfaces/create-email-options.interface.ts
```diff
@@ -79,6 +79,13 @@ interface CreateEmailBaseOptions {
    * @link https://resend.com/docs/api-reference/emails/send-email#body-parameters
    */
   to: string | string[];
+  /**
+   * Schedule the email to be sent later. The time format accepts UTC or local time with a timezone offset.
+   * The time should be in ISO 8601 fAormat (e.g: 2024-08-05T11:52:01.858Z).
```
AntoniosBarotsis: Small typo in `fAormat`, should be `format` 😅

AntoniosBarotsis (changes requested):

src/emails/interfaces/get-email-options.interface.ts
```diff
@@ -25,3 +25,12 @@ export interface GetEmailResponse {
...
 }
+
+export interface CancelScheduleResponse {
+  data: CancelScheduleResponseSuccess | null;
+  error: ErrorResponse | null;
+}
+
+export interface CancelScheduleResponseSuccess {
+  email_id: string;
+}
```
bukinoshita: We need to fix the returned here, it should match the `DELETE domain/:domain_id` for example

felipevolpone: Hm I'm not sure if this is the same. Deleting a domain is different than removing the schedule. If it was to have the same return, it would be something like this:
```
  id: string;
  object: 'schedule';
  deleted: boolean;
```

which looks weird, don't you think?

bukinoshita: That's correct.

Why does it look weird?

felipevolpone: because `schedule` isn't really an object, but ok I don't have strong opinion on this. I'll push the changes

bukinoshita: Got it, should it be `emails`?

src/emails/emails.ts
```diff
@@ -47,4 +49,11 @@ export class Emails {
     );
     return data;
   }
+
+  async cancelSchedule(id: string): Promise<CancelScheduleResponse> {
```
bukinoshita, line 53: the usage should be `resend.email.schedule.remove(id)` instead of `resend.emails.cancelSchedule(id)`

You probably can add this to the class

```
readonly schedule: Schedule;

constructor(private readonly resend: Resend) {
    this.schedule = new Schedule(resend);
}
```

and then define the `Schedule` class similar to the `email`

src/emails/interfaces/create-email-options.interface.ts
```diff
@@ -79,6 +79,13 @@ interface CreateEmailBaseOptions {
    * @link https://resend.com/docs/api-reference/emails/send-email#body-parameters
    */
   to: string | string[];
+  /**
+   * Schedule the email to be sent later. The time format accepts UTC or local time with a timezone offset.
+   * The time should be in ISO 8601 format (e.g: 2024-08-05T11:52:01.858Z).
+   *
+   * @link https://resend.com/docs/api-reference/emails/send-email#body-parameters
+   */
+  schedule_at: string;
```
zenorocha: `schedule_at` → `scheduled_at`

zenorocha:
```suggestion
  scheduled_at: string;
```

src/emails/interfaces/create-email-options.interface.ts
```diff
@@ -79,6 +79,13 @@ interface CreateEmailBaseOptions {
    * @link https://resend.com/docs/api-reference/emails/send-email#body-parameters
    */
   to: string | string[];
+  /**
+   * Schedule the email to be sent later. The time format accepts UTC or local time with a timezone offset.
+   * The time should be in ISO 8601 format (e.g: 2024-08-05T11:52:01.858Z).
```
zenorocha:
```suggestion
   * Schedule email to be sent later.
   * The date should be in ISO 8601 format (e.g: 2024-08-05T11:52:01.858Z).
```

zenorocha (changes requested):

src/emails/emails.ts
```diff
@@ -47,4 +51,11 @@ export class Emails {
     );
     return data;
   }
+
+  async cancel(id: string): Promise<CancelEmailResponse> {
+    const data = await this.resend.post<CancelEmailResponseSuccess>(
+      `/email/${id}/cancel`,
```
bukinoshita:
```suggestion
      `/emails/${id}/cancel`,
```

bukinoshita (approved):

zenorocha (approved):

PR #436, opened 2024-10-05
feat: Improve biome config

eugene-yaroslavtsev:
- use schema from node_modules - makes biome schema, by definition, always be in sync with installed biome version
- use .jsonc to support comments
- standardize to single quotes for JSX as well
- only run checks against files changed relative to main branch
- inherit ignored files from .gitignore
- upgrade usage of soon-to-be deprecated `--fix` command


bukinoshita:
> * use schema from node_modules - makes biome schema, by definition, always be in sync with installed biome version
> * use .jsonc to support comments
> * standardize to single quotes for JSX as well
> * only run checks against files changed relative to main branch
> * inherit ignored files from .gitignore
> * upgrade usage of soon-to-be deprecated `--fix` command

Nice, this is great!

Issue #369, opened 2024-06-07
Have the text property automatically filled when the react property is used when sending emails.


ValentinH:
Do you recommend to always send a text version of an email along with the html version? If yes, could we have the SDK automatically set the `text` property when we use the `react` one (doing a second render with the `plainText: true` option)?
- https://github.com/resend/resend-node/blob/canary/src/emails/emails.ts#L30
- https://github.com/resend/resend-node/blob/canary/src/batch/batch.ts#L27

bukinoshita:
> Do you recommend to always send a text version of an email along with the html version? If yes, could we have the SDK automatically set the `text` property when we use the `react` one (doing a second render with the `plainText: true` option)?
> 
> * https://github.com/resend/resend-node/blob/canary/src/emails/emails.ts#L30
> * https://github.com/resend/resend-node/blob/canary/src/batch/batch.ts#L27

Yes, having the text version of the email is usually recommended.
That's a great suggestion, ideally this should be on API side so user never needs to send the text version doesn't matter which SDK they are using. Doing on SDK side with React Email is easier though

We will explore more

jessethomson:
It seems this still hasn't been implemented.

In the meantime, is this how you all would recommend working around this?
```jsx
import { Resend } from 'resend';
import { render } from '@react-email/render';

const resend = new Resend('re_123456789');

async function sendEmail() {
    const jsx = <MyTemplate />;
    const text = await render(jsx, {
        plainText: true,
    });

    await resend.emails.send({
        from: 'Acme <onboarding@resend.dev',
        to: ['delivered@resend.dev'],
        subject: 'hello world',
        react: jsx,
        text,
    });
}
```

gabrielmfern:
Closing as a duplicate of #451

isabellaaquino:
We just released a fix for this on our API side! We will autogenerate the plain text version if not included. Thanks for reporting!

sommeeeer:
Thank you @isabellaaquino !

PR #469, opened 2025-02-22
feature(email): send email with react component will now render text aswell

### LINKED ISSUE #451: Feature suggestion: Send plain text by default when react email component is provided
I noticed a warning in Resend's insights feature that my nicely styled react emails weren't sending plaintext in addition to the formatted message:

image

I don't know what gave me this perception, but I assumed that resend would automatically generate a plaintext message when a react-email component was supplied to generate an email message.

My suggestion is that plaintext generation should be the _default_ behavior.

For now, I've accomplished this [on my personal site here](https://github.com/mbifulco/blog/pull/901/files#diff-c77a0b10999bb78de544210ec44bf4e497299898622581ea6b6654ef31228ac6) -

```tsx

  import { render } from '@react-email/render';
  import { Resend } from 'resend';

  import { WelcomeEmail } from './templates/WelcomeEmail';

  const resend = new Resend(env.RESEND_API_KEY);

  export const sendWelcomeEmail = async ({
    return;
  }
  
  // this component is made using react-email
  const WelcomeMessage = <WelcomeEmail firstName={firstName} />;

  // send the email
  const { data, error } = await resend.emails.send({
    from: 'Example <hello@example.com>',
    to: [email],
    subject: "Welcome message subject line",
    react: WelcomeMessage,                                    // react component
    text: await render(WelcomeMessage, { plainText: true }),  // plaintext 
  });
```

The pattern I'm using for generating plaintext is derived from https://github.com/resend/react-email/pull/320

sommeeeer:
Closes #451 and #369 

When you pass a React component to either `resend.emails.send()` or `resend.batch.send()` we want it to automatically send the text version aswell. This PR solves that. 

````ts  
const email = await resend.emails.send({
    from: 'test <test@domain.com>',
    to: 'user@domain.com',
    subject: 'hey',
    react: <Email url="https://example.com" />,
    // text: will be automatically rendered here
  });```

### CODE REVIEW

**Review by ThatOneCalculator — APPROVED** [review]
Tested & works!

File: src/batch/batch.ts [hunk 1]
```diff
@@ -38,8 +41,13 @@ export class Batch {
             );
           }
         }
-
-        email.html = await this.renderAsync(email.react as React.ReactElement);
+        const reactElement = email.react as React.ReactElement;
+        email.html = await this.renderAsync(reactElement);
+        if (email.text === undefined) {
```
bukinoshita:
```suggestion
        if (!email.text) {
```

src/emails/emails.ts
```diff
@@ -48,9 +51,13 @@ export class Emails {
         }
       }
 
-      payload.html = await this.renderAsync(
-        payload.react as React.ReactElement,
-      );
+      const reactElement = payload.react as React.ReactElement;
+      payload.html = await this.renderAsync(reactElement);
+      if (payload.text === undefined) {
```
bukinoshita:
```suggestion
      if (!payload.text) {
```

sommeeeer:
I had to add `NODE_OPTIONS=--experimental-vm-modules` in front of the test scripts in `package.json`, else Jest would error on the dynamic import here:
https://github.com/resend/resend-node/blob/0ce0884ccabf01bc3258ba58cd00ff3992c8cc1d/src/emails/emails.ts#L42

The error was: `TypeError: A dynamic import callback was invoked without --experimental-vm-modules`. Its related to Jest not supporting ESM out of the box.

ThatOneCalculator:
Not sure why you're requesting a review from me directly – I'm not involved in the development of this package, but I did use your changes in my application

ThatOneCalculator:
@bukinoshita @sommeeeer anything preventing this from being merged? Already been using this patch in production (on Bun) and it works as expected.

gabrielmfern:
Thank you so much for the contributo, but we've actually added this automatic conversion on the API, so we don't need to do it in the SDK anymore

PR #471, opened 2025-03-15
chore: document the response object in the examples

mcwhittemore:
I found myself wondering what was returned by `resend.emails.send` and ended up using the TypeScript docs and trial and error to figure it out. The docs are pretty light, so I took a lightweight approach, adding a console.log to the examples that explains what just happened, with the ID in the content. I'd love feedback on the approach.


gabrielmfern:
Thank you so much for this, great idea!

PR #474, opened 2025-03-20
feat: add update broadcast

bukinoshita:
this pr adds the patch broadcast method

### DIFF
--- a/src/broadcasts/broadcasts.ts
+++ b/src/broadcasts/broadcasts.ts
@@ -21,6 +21,11 @@ import type {
   SendBroadcastResponse,
   SendBroadcastResponseSuccess,
 } from './interfaces/send-broadcast-options.interface';
+import type {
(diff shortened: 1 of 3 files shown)


src/broadcasts/broadcasts.ts
```diff
@@ -96,4 +101,24 @@ export class Broadcasts {
...
+      `/broadcasts/${id}`,
+      {
+        name: payload.name,
+        audience_id: payload.audienceId,
+        from: payload.from,
+        html: payload.html,
+        text: payload.text,
+        subject: payload.subject,
+        reply_to: payload.replyTo,
+        preview_text: payload.previewText,
```
felipevolpone, line 119: what about the schedule_at field?

bukinoshita: we don't have that in the API

felipevolpone (approved):

PR #480, opened 2025-03-26
fix: prevent `prettier` version conflict in pnpm installations

AjayPawarExist:
## Summary  
When installing `resend` in a Next.js project using `pnpm`, an error occurs due to a version mismatch between `prettier` versions. The project may use `prettier@3.5.3`, while `@react-email/render` depends on `prettier@3.4.2`, causing conflicts in PnP environments.  

## Fix Implemented  
- Updated the `prettier` version in dependencies to match newer versions used in modern projects.  
- Ensured compatibility to prevent external package resolution errors in `pnpm`.  

## Why This Fix is Needed  
Even if dependencies are updated, mismatches between project-level and package-level versions can cause issues, especially in strict environments like `pnpm` (Plug’n’Play). This update ensures `resend` remains compatible with modern setups.  

## Error Screenshot  
image

## Resolution  
With this update, `resend` can be installed without conflicts, ensuring smoother integration in projects using `pnpm`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "homepage": "https://github.com/resendlabs/resend-node#readme",
   "dependencies": {
-    "@react-email/render": "1.0.5"
+    "@react-email/render": "1.0.6"
   },
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       '@react-email/render':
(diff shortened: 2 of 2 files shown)


bdrtsky:
Why this ignored? How can I use resend if it's flooding my console with conflict messages?

bukinoshita:
@AjayPawarExist thanks a lot for the fix, can you fix the conflict?

AjayPawarExist:
Thanks! Actually, Renovate[bot] fixed the issue in its own PR and auto-merged it before mine could be merged.

PR #492, opened 2025-04-18
feat: add custom return path support for domains creation

isabellaaquino:
This PR adds support for sending a `customReturnPath` field to domains creation (`resend.domains.create`). It allows users to customize their SPF record host names.

The field is supposed to be optional and correctly parsed to snake case to match our API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "resend",
-  "version": "4.3.0",
+  "version": "4.4.0-canary-0",
   "description": "Node.js library for the Resend API",
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "resend",
-  "version": "4.4.0-canary-0",
+  "version": "4.3.0",
   "description": "Node.js library for the Resend API",
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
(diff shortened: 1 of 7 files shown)


src/common/interfaces/domain-api-options.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface DomainApiOptions {
```
jamescmartinez, line 1: I see this in another part of the PR. Is it a duplicate interface that we could consolidate / DRY up?

isabellaaquino: This interface is indeed very similar to `CreateDomainOptions`, but the difference is that the fields are in snake case to match our API. 
It only exists to type the parsing we are doing to the payload and it's following the pattern used by another similar file `email-api-options.interface.ts`!

WDTY?

jamescmartinez: 🤦‍♂️ my mistake, sorry about that!

PR looks good to me.

jamescmartinez (approved):

src/common/interfaces/domain-api-options.interface.ts
```diff
@@ -0,0 +1,5 @@
...
+  name: string;
+  region?: string;
+  custom_return_path?: string;
```
bukinoshita, line 4: This should be camel cased, only in the API is snake cased

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "resend",
-  "version": "4.3.0",
+  "version": "4.4.0-canary-0",
```
bukinoshita: Can you revert that, we can bump the version once we release since it may include more stuff

bukinoshita (approved):

PR #513, opened 2025-05-02
fix: Add idempotency error codes

abeisleem:
Adds the error codes as per https://resend.com/docs/dashboard/emails/idempotency-keys#possible-responses


joaopcm:
@abeisleem, these changes are now released in the [v4.5.1](https://www.npmjs.com/package/resend/v/4.5.1). Thanks!

Issue #525, opened 2025-05-22
Resend should not have a React Dependency


ConProgramming:

takeitstore:
+1. I got really confused when I saw prettier and @react-email/render in my sveltekit production build

ValiuchenkoVladyslav:
+1. I got really confused when I saw prettier and @react-email/render in my sveltekit production build

mnove:
Same - just noticed in my build as well

gabrielmfern:
Closing as a duplicate of #309

treeder:
Why is this closed?  I see react dependencies all over the place, even in canary branch: https://github.com/resend/resend-node/blob/canary/src/emails/emails.ts

Issue #503, opened 2025-04-26
Resend require react to be used


flixyudh:
I just read the [CF documentation about to send email](https://developers.cloudflare.com/workers/tutorials/send-emails-with-resend/) and found resend.

when I follow the doc, I got an error 

```✘ [ERROR] Build failed with 3 errors:

  ✘ [ERROR] Could not resolve "react"
  
      node_modules/@react-email/render/dist/browser/index.mjs:43:25:
        43 │ import { Suspense } from "react";
           ╵                          ~~~~~~~
  
    You can mark the path "react" as external to exclude it from the bundle, which will remove this
  error and leave the unresolved path in the bundle.
  
  
  ✘ [ERROR] Could not resolve "react/jsx-runtime"
  
      node_modules/@react-email/render/dist/browser/index.mjs:121:20:
        121 │ import { jsx } from "react/jsx-runtime";
            ╵                     ~~~~~~~~~~~~~~~~~~~
  
    You can mark the path "react/jsx-runtime" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle.
  
  
  ✘ [ERROR] Could not resolve "react-dom/server"
  
      node_modules/@react-email/render/dist/browser/index.mjs:156:38:
        156 │   const reactDOMServer = yield import("react-dom/server");
            ╵                                       ~~~~~~~~~~~~~~~~~~
  
    You can mark the path "react-dom/server" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle. You can also add ".catch()" here to
  handle this failure at run-time instead of bundle-time.
```

I know the error appear because I'm not installing `react`, but why I need to install `react` while my project is full of rest API.

is it possible for resend team to create a node SDK without any dependencies?

gabrielmfern:
Closing as a duplicate of #309

treeder:
@gabrielmfern none of these are duplicates of #309. 309 is for @react-email/render dependency, not react itself.

Issue #406, opened 2024-08-03
React declaration not found in Fastify (Node.js) app


shidoxo:
```
node_modules/resend/dist/index.d.ts:1:24 - error TS7016: Could not find a declaration file for module 'react'. 'C:/Users/<redacted>/<redacted>/<redacted>/<redacted>/<redacted>/node_modules/react/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`

1 import * as React from 'react';

Found 1 error in node_modules/resend/dist/index.d.ts:1
```

Why do i need React at all in my NodeJS server? If there's React, why is declaration file not provided?

mcfarljw:
I'm seeing the same same issue. The reason is because there is a react dependency so for those not using react it's causing typing errors.

https://github.com/resend/resend-node/blob/c69c13f582314011281ba2b667a9645978f43839/package.json#L45C5-L45C36 this should not be a required dependency as many aren't using react.

chalscc:
Experiencing the same issue, adding the following line on tsconfig.json solved the issue for me, without needing to install react types

On `compilerOptions`

`    "skipLibCheck": true`

gabrielmfern:
Recommended to use `"skipLibCheck": true` here. There's no way for us to get around this while still supporting React for the users that have React.

Issue #353, opened 2024-05-09
Why Resend need React to work in Express??


AlejandroMora15:
I am develop an API in express with Typescript, and I am using Resend to send emails. I followed the instructions of the official doc page: https://resend.com/docs/send-with-express.

In local works perfectly. However, when I try to deploy in prod creating the build, I get this error:

`
node_modules/resend/dist/index.d.ts(1,24): error TS7016: Could not find a declaration file for module 'react'. '/opt/render/project/src/node_modules/react/index.js' implicitly has an 'any' type.
`

I cannot find any solution. Why Resend need React to work in express?

SirBenedick:
@AlejandroMora15 you don't need react to send an email with resend on an express server. 

Did you use react components in the email body? 

I did not need to install react for using the example 
`  const { data, error } = await resend.emails.send({
    from: "Acme <onboarding@resend.dev>",
    to: ["delivered@resend.dev"],
    subject: "hello world",
    html: "<strong>it works!</strong>",
  });`

AlejandroMora15:
The problem is when I install Resend in my express BE, it also install React!

michalzarsm:
Same issue working with Fastify

tylerclark:
Same issue using fastify too

SirBenedick:
> The problem is when I install Resend in my express BE, it also install React!
@AlejandroMora15 you might have react as a peer dependency of a package inside your express project. 

If so you might need to import react without using it directly:
``` typescript
// @ts-ignore-line
import React from "react";
```

tylerclark:
If you are importing a frontend JavaScript framework into your backend Node project, something is wrong.

AlejandroMora15:
> If you are importing a frontend JavaScript framework into your backend Node project, something is wrong.

First, you must improve your reading comprehension. 

What I’m explaining is that when I install Resend, React gets installed as well! I don’t need React in the backend, but Resend installs it automatically, and when I deploy, it throws that error.

ViaAnthroposBenevolentia:
Hey! Had this issue on my Fastify server too. Fortunately, turns out you only need to install react types to a **dev** dependencies (pnpm example):
```bash
pnpm install -D @types/react
```

Then build to check if compiler would complain. My build command is:
` "build": "tsc && tsc-alias",`

And my tsconfig.json:
```JSON
{
    "compilerOptions": {
        "target": "ES2022",
        "module": "NodeNext",
        "moduleResolution": "NodeNext",
[...]

PR #527, opened 2025-06-06
fix!: make @react-email/render an optional peer dependency

emiliosheinz:
## Summary

Several Resend users have been complaining about unexpectedly seeing `@react-email/render` and `react` in their builds. After analysis, it was found that this was happening because `@react-email/render` was listed as a dependency of resend-node.

To fix this, I decided to make `@react-email/render` an optional peer dependency. This way, it won’t be automatically installed when `resend-node` is installed.

No functionality should be affected, as `@react-email/render` is dynamically imported throughout the codebase. However, this is a breaking change for users who use the react parameter to send emails with resend-node, as they will now need to install `@react-email/render` themselves.

## Additional information

I'm opening this PR to start the discussion around the topic and get approval from the maintainers to move forward with that effort, as further changes will have to be made to the docs, mentioning that now users who want to send emails using React will have to install `@react-email/render` themselves.

Additionally, I believe that, in an ideal world, `resend-node` should not depend directly on `@react-email/render`. Instead, it should rely on an abstraction, such as a function passed into the Resend constructor, that handles the conversion. This would decouple `resend-node` from any specific rendering implementation. Alternatively, the package could be simplified to only accept `html` or `text`, allowing users to decide how they want to convert React components into email content using any library they prefer. This approach would make `resend-node` more flexible and extensible.

However, implementing such a change would require a significant architectural refactor, so it’s not suitable as a quick fix.

## Potentially related open issues

[Resend should not have a React Dependency](https://github.com/resend/resend-node/issues/525)
[Resend require react to be used](https://github.com/resend/resend-node/issues/503)
[React declaration not found in Fastify (Node.js) app](https://github.com/resend/resend-node/issues/406)
[Why Resend need React to work in Express??](https://github.com/resend/resend-node/issues/353)
[Resend SDK design](https://github.com/resend/resend-node/issues/309)
[Remove dependency on react package to make it more lightweight](https://github.com/resend/resend-node/issues/243)

Happy to hear your thoughts and feedback :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,13 @@
     "url": "https://github.com/resendlabs/resend-node/issues"
   },
   "homepage": "https://github.com/resendlabs/resend-node#readme",
-  "dependencies": {
+  "peerDependencies": {
     "@react-email/render": "1.1.2"
+  },
+  "peerDependenciesMeta": {
+    "@react-email/render": {
+      "optional": true
+    }
   },
   "devDependencies": {
     "@biomejs/biome": "1.9.4",


vcapretz:
really glad you opened this open! I actually didn't see it on time and opened a new one #532 😓 I'll iterate in there and get it merged 🙏

PR #528, opened 2025-06-06
refactor: remove deprecated usage of render async

emiliosheinz:
This refactor removes the usage of the deprecated `renderAsync` function from `@react-email/render`, replacing it with the updated and recommended `render function`. While both methods provide the same core functionality, `render` offers a clearer and more consistent naming.

This change maintains the existing behavior while improving code clarity and future compatibility, reducing technical debt, and ensuring alignment with the latest library standards.

### DIFF
--- a/src/batch/batch.ts
+++ b/src/batch/batch.ts
@@ -10,7 +10,7 @@ import type {
 } from './interfaces/create-batch-options.interface';
 
 export class Batch {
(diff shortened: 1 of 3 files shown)


gabrielmfern:
Sorry for not merging this before, but this was also done as part of #620, so closing this, but thank you so much for contributing!

Issue #467, opened 2025-02-04
Edge Runtime Warning Due to Unconditional @react-email/render Import


mzubair481:
## Description
When using the Resend SDK in a Next.js 15 project, we get an Edge Runtime warning even when not using React components for emails. The warning appears because the SDK unconditionally imports `@react-email/render`, even when only using the `text` option for emails.

## Error Message
```
A Node.js API is used (MessageChannel at line: 126) which is not supported in the Edge Runtime.

Import trace for requested module:
./node_modules/next/dist/compiled/react-dom/cjs/react-dom-server.browser.production.js
./node_modules/next/dist/compiled/react-dom/server.browser.js
./node_modules/resend/node_modules/@react-email/render/dist/browser/index.mjs
./node_modules/resend/dist/index.mjs
./src/services/email.service.ts
```

## Steps to Reproduce
1. Create a Next.js 15 project
2. Install `resend`
3. Use the SDK to send a plain text email:
   ```typescript
   await resend.emails.send({
     from: "...",
     to: "...",
     subject: "...",
     text: "Hello world"  // Only using text, no React
   });
   ```
4. Run `next build`

## Expected Behavior
When using only the `text` option (no React components), the SDK should not import `@react-email/render` since it's not needed.

## Actual Behavior
The SDK always imports `@react-email/render`, causing Edge Runtime warnings in Next.js projects even when not using React components.

## Environment
- Next.js: 15.1.5
- Resend SDK: 4.1.1
- Node.js: 22.12.0

emiliosheinz:
Hey @mzubair481, can you please share more information about your issue?

- Are you using Nextjs `app` or `page` router?
- Does the error happen at build or run time?
- Do you have an example repo where I can look at the code?

I tried to reproduce it on my side but had no luck. Maybe I'm missing something.

vcapretz:
this should now be fixed since we are using dynamic imports, let me know if it still happens in the latest version of the SDK.
otherwise we have another improvement coming up to avoid installing the render package completely in #532

krisnaw:
hi @vcapretz this issue still persists on a fresh Next.js project. I'm using "resend": "^6.1.0", and sending the email from API routes as suggested in the documentation.

gabrielmfern:
@krisnaw can you share a reproduction for that?

krisnaw:
Hi @gabrielmfern here are the step to reproduce
1. install next.js `npx create-next-app@latest`
2. install resend `npm install resend`
[...]

Issue #243, opened 2023-10-26
Remove dependency on react package to make it more lightweight


awdr74100:
Hi, first of all thanks for such a great email service, I really like it!

I found that this package is much larger than the packages provided by other email services. Is there any plan to remove the dependency of the `react` package? To make it more lightweight and less stressful for the integration of other frameworks. Or is this package specific to integrating `react`?

I know it's possible to simply use the rest api, but then won't get the benefits of types.

azizsafudin:
I believe `react` is included due to `react-email/render`, which is unnecessary for such an SDK. I believe they should be able to make this a standalone client SDK, without `react-email/render`.

Commenting because I am facing issues due to `react-email/render` being included in the package.

vcapretz:
hey there, `react` itself is now marked as a `peerDependency` for the `react-email/render` package, so it should already alleviate a bit the size of the SDK package itself, but I do agree that not depending on `react-email/render` would be even better.

unfortunately that would put a burden on our users so they have to remember to add extra dependencies separately when they want to use the `react` prop in the SDK.

do you think the size as is it's in a bad state? we can try looking into minimizing the dependencies for `react-email/render` a little bit, but I wouldn't say that's a major concern

azizsafudin:
Hi, it’s been awhile since I left my comment, I’ve forgotten some of the context but for us the size wasn’t the issue. I was just not able to use it in certain serverless environments. I think this coupling is unnecessary for a resend package. If the concern is that users have to add extra deps, then that’s the proper way to do it, along with docs explaining how to get it working. FYI, this was critical enough that we ended up using Postmark instead.

kedarv:
++ echoing/bumping this thread, having to bundle in react and react-dom for nodeJS applications is a massive drawback of this library. I may end up swapping the client implementation out for a generic SMTP library instead, which is unfortunate because it removes most/all of the DX facing gains by using a native sdk

h4ckedneko:
Hello, I've been using Resend for my applications because its DX is very good. Recently I've been using Cloudflare Workers too where your Worker code size matters. Because this library depends on React, even though I don't use React at all in my email templates. The Worker runtime parses those unused code and degrading the start-up time of my Worker. I believe removing React as a required dependency and making it an optional pluggable dependency is better.

![image](https://github.com/user-attachments/assets/36917831-c69c-4f2c-93e2-1b56048c42ce)
[...]

PR #532, opened 2025-06-12
feat: make `@react-email/render` an optional _peerDependency_

vcapretz:
## 🚨🚨 note: this is a breaking change if you're using the `react` prop. make sure you have either `@react-email/components` or `@react-email/render` installed

we previously introduced dynamic imports for the `@react-email/render` package to alleviate issues in different node  environments, but we would still install the package, causing issues and frustrations: #525 #503 #467 #406 #353 #243 

this PR finally proposes a solution by introducing an optional peerDependency.

note that most package managers nowadays do install `peerDependencies` by default, so we have to explicitly set the `@react-email/render` one as optional.

**another note:** this should be a major version bump, because for people who do use the `react` feature they will have to explicitly install the package, otherwise they will see errors being thrown.

| before | after | 
| ----- | ----- | 
| CleanShot 2025-06-12 at 13 37 40@2x |  CleanShot 2025-06-12 at 13 38 46@2x |
| CleanShot 2025-06-12 at 13 38 00@2x | CleanShot 2025-06-12 at 13 40 18@2x |

> if you have any other concerns or suggestions, please drop them below! 😄


package.json
```diff
@@ -41,9 +41,14 @@
     "url": "https://github.com/resendlabs/resend-node/issues"
   },
   "homepage": "https://github.com/resendlabs/resend-node#readme",
-  "dependencies": {
+  "peerDependencies": {
     "@react-email/render": "1.1.2"
```
gabrielmfern: did you also test this with just `@react-email/components` installed with `bun`, `yarn`, `yarn modern`, `npm` & `pnpm`?

would be cool to have https://github.com/stackblitz-labs/pkg.pr.new integration to help with this

vcapretz:
> Overall looking good, but the imports of `render` are still messy - using `renderAsync`, storing in a property of the class, would definitely recommend a refactor there

agreed! I'll do it separately tho

emiliosheinz:
Great improvement @vcapretz! Thank you for looking into it 🚀 

I just want to circle back on something I mentioned in #527 that I believe would be the best way forward and long-term solution.

I personally believe that `resend-node` should not depend directly on `@react-email/render`. Instead, it should rely on an abstraction, such as a function passed into the Resend constructor, or it could be simplified to only accept `html` or `text`, allowing users to decide how they want to convert React components into email content using any library they prefer.

Here is an example of the latter:

```jsx
import { render } from '@react-email/render';

import EmailTemplate from '../components/EmailTemplate';

await resend.emails.send({
  from: 'you@example.com',
  to: 'user@gmail.com',
  replyTo: 'you@example.com',
  subject: 'hello world',
  // This would completely decouple both libraries 
  html: await render(<EmailTemplate firstName="John" product="MyApp" />),
});
```

This approach would make `resend-node` more flexible and extensible, decoupling `resend-node` from any specific rendering implementation. 

I’d be happy to discuss this further and collaborate if you’re interested.

vcapretz:
I see your point @emiliosheinz, and I also thought about it.

Indeed completely getting rid of the `react` option would be the clearest path as they both get completely separated, but I wonder if having it in the first place is some small feature that will delight users and improve the experience for the ones that need it

emiliosheinz:
@vcapretz Maybe we could open an RFC issue in this repo asking for community feedback about the preferred path forward?

molvqingtai:
Even though I didn't use @react-email/render, it still occupies several hundred KB, which is unacceptable for a project running in a edge/worker.
Google Chrome 2025-07-08 22 41 54

molvqingtai:
Temporary solution:
```js
  webpack: (config) => {
    if (Array.isArray(config.externals)) {
      config.externals.push({ '@react-email/render': {} })
      config.externals.push({ '@react-email/components': {} })
    }
    return config
  }
```

emiliosheinz:
Hey @vcapretz, is there anything I can do to help moving this forward? It looks like a pain for several users.

PR #535, opened 2025-06-13
feat: update types to allow type narrowing

### LINKED ISSUE #356: Type annotations improvement
```javascript
const { data, error } = await resend.emails.send({
    from: 'Acme <onboarding@resend.dev>',
    to: ['delivered@resend.dev'],
    subject: 'Hello World',
    html: '<strong>It works!</strong>',
});

if (error) {
    return console.error({ error });
}

console.log({ id: data.id });
```

taking an example from https://resend.com/docs/send-with-nodejs, If you access id from data directly, `ts` will complain that `'data' is possibly 'null'`, and that's because of return type:
```javascript
 async fetchRequest<T>(
    path: string,
    options = {},
  ): Promise<{ data: T | null; error: ErrorResponse | null }>
```

and I think it's not accurate because from the code we can know that data can not be null when error is null, and it can be changed like this:
```javascript
export type SuccessResponse<T> = {
  data: T;
  error: null;
};

export type FailureResponse = {
  error: ErrorResponse
  data: null;
}

export type Result<T> = SuccessResponse<T> | FailureResponse;
async fetchRequest<T>(
    path: string,
    options = {},
 ): Promise<Result<T>>
```

vcapretz:
closes #356

previously we had the response schemas defining both `data` and `error` to be either the value or _null_. but as correctly pointed in the issue, we always have _either/or_

by updating the type definitions to something like:
```
// from this:
type Response = { data: MyData | null; error: MyError | null }

// to this:
type Response = { data: MyData; error: null } | { data: null; error: MyError }
```

we can leverage TypeScript type narrowing in the code and have better types 🎉 

**note the null in here:**
CleanShot 2025-06-13 at 16 42 14@2x

**much better:**
CleanShot 2025-06-13 at 16 51 10@2x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 node_modules
 dist
-.yarn/*
-!.yarn/releases
+build
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "resend",
-  "version": "4.6.0",
+  "version": "4.7.0-canary.0",
   "description": "Node.js library for the Resend API",
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
--- a/src/api-keys/interfaces/create-api-key-options.interface.ts
+++ b/src/api-keys/interfaces/create-api-key-options.interface.ts
@@ -14,7 +14,12 @@ export interface CreateApiKeyResponseSuccess {
   id: string;
 }
 
-export interface CreateApiKeyResponse {
-  data: CreateApiKeyResponseSuccess | null;
-  error: ErrorResponse | null;
-}
+export type CreateApiKeyResponse =
+  | {
+      data: CreateApiKeyResponseSuccess;
+      error: null;
+    }
+  | {
+      data: null;
+      error: ErrorResponse;
+    };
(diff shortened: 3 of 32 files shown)


joaopcm (approved):

tadeumaia:
Should changing types be a major bump on semver as it causes breaking code?

Works perfectly fine on 4.6 but breaks on 4.7

```
  it("should throw error if Resend API fails", async () => {
    // Arrange
    const mockResendResponse: ResendEmailResponse = {
      data: null,
      error: {
        message: "Failed to send email",
        name: "invalid_parameter",
      },
    };
    vi.mocked(resend.emails.send).mockResolvedValue(mockResendResponse);

    // Act & Assert
    await expect(sendInvitationEmail(mockEmailData)).rejects.toThrow("Failed to send invitation email");
  });
```

```
tests/unit/lib/email/index.test.ts:499:53 - error TS2345: Argument of type 'ResendEmailResponse' is not assignable to parameter of type 'CreateEmailResponse'.
  Type 'ResendEmailResponse' is not assignable to type '{ data: null; error: ErrorResponse; }'.
    Types of property 'data' are incompatible.
      Type '{ id: string; } | null' is not assignable to type 'null'.
        Type '{ id: string; }' is not assignable to type 'null'.

499     vi.mocked(resend.emails.send).mockResolvedValue(mockResendResponse);
```

PR #539, opened 2025-06-25
fix: update info about runtime JSX rendering

cassiozen:
When using the `react` prop, we suggest calling React components as functions when JSX transpilation isn't available (e.g., `EmailTemplate({firstName: "John", product: "MyApp"})`). However, this approach has several limitations:

- Components that use hooks may not work correctly when called directly as functions
- React 19 introduces changes in component types (specifically `React.FC`) that will trigger errors 

Solution: Recommend using `jsx` from `react/jsx-runtime` instead. This function is what build tools like Babel actually use to transform JSX at compile time, making it the most reliable alternative when JSX syntax isn't available.

This change updates the README to recommend the jsx function approach.


gabrielmfern (approved):
LGTM!

PR #540, opened 2025-06-26
chore(deps): upgrade biome

emiliosheinz:
This PR upgrades `@biomejs/biome` from version `1.9.4` to `2.0.5`.

### Changes made:

- Bumped the package version.
- Ran the official migration tool:  
  ```bash
  npx @biomejs/biome migrate --write
  ```
- Addressed all linting and formatting issues introduced by the new version.
- Ensured the codebase is fully compatible with Biome 2.0.5.

###  Notes:
The [Original Renovate PR](https://github.com/resend/resend-node/pull/537) failed due to linting issues. This PR supersedes it by including all required fixes. Therefore, the pipelines should now pass successfully.


emiliosheinz:
@gabrielmfern there seems to be something wrong with the pipeline YAML files. Do you want me to look into it?

gabrielmfern:
@emiliosheinz you shouldn't have to, it's a problem on how we have it configured. Basically only works easily for us, and for contributors it's terrible right now. I'm discussing with the team how we can make something better

PR #554, opened 2025-08-01
feat: add `inline_content_id` to attachments

Cisneiros:
Adds support for the optional `inline_content_id` property in email attachments.


src/common/interfaces/email-api-options.interface.ts
```diff
@@ -14,5 +21,5 @@ export interface EmailApiOptions {
   reply_to?: string | string[];
   scheduled_at?: string;
   tags?: Tag[];
-  attachments?: Attachment[];
```
rehanvdm, line 17: Will this be a breaking change?

Cisneiros: No, this updates the API-facing schema but not the user-facing schema. This function adapts the SDK user input to the format expected by the REST API.

rehanvdm: Stamped

PR #555, opened 2025-08-01
chore: use `render` as a `devDependency`

gabrielmfern:
Continuing on the work of #527.

After testing out the `5.0.0-canary.0` on an existing project I noticed that the approach of using peer dependencies made it so that if the user already had installed the previous version, consequently having `@react-email/render` in their lockfile, and then updated, it would keep `@react-email/render` in the lockfile for pnpm, npm and bun. This would not happen for new installs, though.

If we were to keep it as a peer dependency, we'd need to tell users to either remove things from their lockfile manually, or recreate their lockfile (which they should probably not do).

This pull requests changes it so that we use `devDependencies` instead, this avoids that problem and will reduce the installed dependencies after upgrading.

I tested this with a package prefixed with my name, you can compare installing `gabriel-resend-node@4.7.0`, and then installing `gabriel-resend-node@5.0.0-canary.0` and notice that the lockfile does not include `@react-email/render` after the version change. Here's what you will find with pnpm:

| 4.7.0 | 5.0.0-canary.0 |
|--------|--------|
| image | image |
| image | image |

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "homepage": "https://github.com/resendlabs/resend-node#readme",
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
-    "@react-email/render": "1.1.0",
+    "@react-email/render": "1.1.2",
     "@types/jest": "29.5.14",
     "@types/node": "18.19.86",
     "@types/react": "19.1.2",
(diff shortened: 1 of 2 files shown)


gabrielmfern:
We're instead opting into using the `peerDependency` and telling users to uninstall and then install the sdk if `@react-email/render` remains installed in the lockfile. This is a temporary problem, not a permanent one, so it's not worth it to go for `devDependencies`.

PR #562, opened 2025-08-08
feat: rate limiting information for all responses

### LINKED ISSUE #189: The rate limit headers aren't returned to users
If there's a rate limit error, it would be good if the rate limit headers are added to the error object. Currently it's not possible to get them when using the SDK.

gabrielmfern:
Closes #189.

Adds a type to use on all responses to avoid repeating the definition for it, and add a `rateLimiting` field there that gives access to the `ratelimit-limit`, `ratelimit-reset` and `ratelimit-remaining` headers.

The actual changes to use the new `Response` type were made by claudinho, so take that portion with a grain of salt

Here's how an example of the API would feel like:

```typescript
const response = await resend.emails.send(/* ... */);

//  limit: number;
//  remainingRequests: number;
//  shouldResetAfter: number;
if (response.rateLimiting?.remainingRequests === 0) {
	setTimeout(() => {
		// ...continue making requests...
	}, response.rateLimiting.shouldResetAfter * 1000);
}
```

Unfortunately, there are a few cases that are errors that happen on the client and not on the server, so if `error` is not null, `rateLimiting` can be null. Ideally the user would not need to handle that, but seems like it's unavoidable.

Additionally, this also adds in a `retryAfter` property to the ErrorResponse when the `name` is 'rate_limit_exceeded' taken from the `retry-after` header that only comes on 429 responses. It feels like this:

```typescript
const response = resend.emails.send(/* ... */);

if (response.error?.name === 'rate_limit_exceeded') {
	setTimeout(() => {
		/* ...code for retrying... */
	}, 1000 * response.error.retryAfter);
}
```

### DIFF
--- a/src/rate-limiting.ts
+++ b/src/rate-limiting.ts
@@ -1,6 +1,19 @@
 export type RateLimit = {
+  /**
+    * The maximum amount of requests that can be made in the time window of {@link shouldResetAfter}.
+    */
   limit: number;
+  /**
+    * The amount of requests that can still be made before hitting {@link RateLimit.limit}.
+    *
+    * Resets after the seconds in {@link RateLimit.shouldResetAfter} go by.
+    */
   remainingRequests: number;
+  /**
+    * The number of seconds after which the rate limiting will reset, 
+    * and {@link RateLimit.remainingRequests} goes back to the value of 
+    * {@link RateLimit.limit}.
+    */
   shouldResetAfter: number;
 };
 
--- a/src/rate-limiting.ts
+++ b/src/rate-limiting.ts
@@ -1,19 +1,24 @@
+// this is used in the jsdoc for `shouldResetAfter`
+import type { Response } from './interfaces';
+
 export type RateLimit = {
   /**
-    * The maximum amount of requests that can be made in the time window of {@link shouldResetAfter}.
-    */
+   * The maximum amount of requests that can be made in the time window of {@link shouldResetAfter}.
+   */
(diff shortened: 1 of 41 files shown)


src/rate-limiting.ts
```diff
@@ -0,0 +1,45 @@
...
+export function parseRateLimit(headers: Headers): RateLimit {
+  const limitHeader = headers.get('ratelimit-limit');
+  const remainingHeader = headers.get('ratelimit-remaining');
+  const resetHeader = headers.get('ratelimit-reset');
+
+  if (!limitHeader || !remainingHeader || !resetHeader) {
+    throw new Error(
+      "The rate limit headers are not present in the response, something must've gone wrong, please email us at support@resend.com",
+    );
+  }
```
joaopcm, line 34: What if we gracefully fail here? This would be a big problem if we change how the rate limit works in the API

gabrielmfern: that function is only called from 

https://github.com/resend/resend-node/blob/983cc6a773ad41f44888d8e64e1a993f7ecfbf28/src/resend.ts#L60-L62

and there it errors gracefully, but with an error message that isn't very clear. we could return something else for when it isn't able to get the rate limiting headers, I also thought of just warning, but my worry with not erroring here is that it could break workflows for users silently

imo the best would be to have tests on the SDK's side that tested things directly with the production API ensuring that this rate limiting is working so it would fail early, instead of doing mocking

PR #569, opened 2025-08-19
feat: add create template method

CarolinaMoraes:
## Description
This PR implements the method for template creation. 
Worth mentioning that: besides HTML, it also supports React, as mentioned in the RFC

To support our current template syntax, and avoid users from ending up with a broken React code, the new.email approach to interpolations will be used ([this approach was suggested here](https://resend.slack.com/archives/C094Q028PK3/p1753889275450259?thread_ts=1753883774.546519&cid=C094Q028PK3))

Which means the code (for someone that will use React) will look something like the following:

Template code
```jsx
<Heading className="text-[24px] font-bold text-gray-900 mb-[16px]">
   Welcome to Your New Beginning, {props.employeeName}
</Heading>
```

SDK call in a node environment
```js
        const template = await resend.templates.create({
            name: "Welcome to Lumon",
            subject: "Welcome to Lumon Industries - Your journey begins now",
            react: React.createElement(LumonWelcomeEmail, {
                employeeName: "{{{employeeName}}}",
            }),
            variables: [
                { key: "employeeName", type: "string", fallback_value: "Mark S." },
            ]
        })
```

### DIFF
--- a/src/templates/interfaces/template.ts
+++ b/src/templates/interfaces/template.ts
@@ -1,0 +1,23 @@
+export interface Template {
+  id: string;
+  name: string;
+  subject?: string;
+  html: string;
+  text?: string;
+  status: 'draft' | 'published' | 'deleted';
+  variables?: TemplateVariable[];
+  alias?: string;
+  from?: string;
+  reply_to?: string | string[];
+  published_at?: string;
+  created_at: string;
+  updated_at: string;
+}
+
+export interface TemplateVariable {
+  key: string;
+  fallback_value?: string;
+  type: 'string' | 'number' | 'boolean';
+  created_at: string;
+  updated_at: string;
+}
--- a/src/templates/templates.spec.ts
+++ b/src/templates/templates.spec.ts
@@ -42,22 +42,22 @@ describe('Templates', () => {
       await expect(
         resend.templates.create(payload),
       ).resolves.toMatchInlineSnapshot(`
-{
-  "data": {
-    "id": "3deaccfb-f47f-440a-8875-ea14b1716b43",
-    "object": "template",
-  },
-  "error": null,
-}
-`);
+    {
+      "data": {
+        "id": "3deaccfb-f47f-440a-8875-ea14b1716b43",
+        "object": "template",
+      },
+      "error": null,
+    }
+    `);
     });
 
     it('creates a template with all optional fields', async () => {
       const payload: CreateTemplateOptions = {
         name: 'Welcome Email',
         subject: 'Welcome to our platform',
-        html: '<h1>Welcome to our platform!</h1>',
-        text: 'Welcome to our platform!',
+        html: '<h1>Welcome to our platform, {{{name}}}!</h1><p>We are excited to have you join {{{company}}}.</p>',
+        text: 'Welcome to our platform, {{{name}}}! We are excited to have you join {{{company}}}.',
         variables: [
           {
             key: 'name',
@@ -91,21 +91,34 @@ describe('Templates', () => {
       await expect(
         resend.templates.create(payload),
       ).resolves.toMatchInlineSnapshot(`
-{
-  "data": {
-    "id": "fd61172c-cafc-40f5-b049-b45947779a29",
-    "object": "template",
-  },
-  "error": null,
-}
-`);
+    {
+      "data": {
+        "id": "fd61172c-cafc-40f5-b049-b45947779a29",
+        "object": "template",
+      },
+      "error": null,
+    }
+    `);
     });
 
     it('creates a template with some optional fields', async () => {
       const payload: CreateTemplateOptions = {
         name: 'Password Reset',
         subject: 'Reset your password',
-        html: '<h1>Reset your password</h1><p>Click the link below to reset your password.</p>',
+        html: '<h1>Reset your password, {{{username}}}</h1><p>Click the link below to reset your password: {{{reset_link}}}</p>',
+        text: 'Reset your password, {{{username}}}. Click here to reset: {{{reset_link}}}',
+        variables: [
+          {
+            key: 'username',
+            fallback_value: 'there',
+            type: 'string',
+          },
+          {
+            key: 'reset_link',
+            fallback_value: 'https://example.com/reset',
+            type: 'string',
+          },
+        ],
         alias: 'password-reset',
         from: 'security@example.com',
       };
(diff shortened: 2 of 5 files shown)


src/templates/interfaces/template.ts
```diff
@@ -0,0 +1,23 @@
+export interface Template {
+  id: string;
+  name: string;
+  subject?: string;
+  html: string;
+  text?: string;
+  status: 'draft' | 'published' | 'deleted';
```
joaopcm: I don't think we'll ever return `status === deleted`, will we?

src/templates/interfaces/template.ts
joaopcm, line 1: At the API level, these fields cannot be `undefined`, as covered by the `?` TS operator. Instead, they can only be `null` or truthy. Is this something we should cover in this PR?

src/templates/templates.spec.ts
```diff
@@ -0,0 +1,735 @@
...
+      const result = resend.templates.create(payload);
+
+      await expect(result).resolves.toMatchInlineSnapshot(`
+    {
+      "data": null,
+      "error": {
+        "message": "Missing \`name\` field",
+        "name": "missing_required_field",
+      },
+    }
```
joaopcm: Shouldn't it include `statusCode` as well?

src/templates/templates.spec.ts
```diff
@@ -0,0 +1,735 @@
...
+
+    it('throws error when missing required html field', async () => {
+      const payload: CreateTemplateOptions = {
+        name: 'Welcome Email',
+        html: '',
+      };
+      const response: ErrorResponse = {
+        name: 'missing_required_field',
+        message: 'Missing `html` field',
+      };
```
joaopcm: Same about status code

src/templates/templates.spec.ts
joaopcm, line 1: I think most of the validation errors described in this test file is just mocking the API response and confirming the error is returned within the `error` object in the function output. Does it make sense to mock multiple different error messages and checking the same response structure? Wouldn't only one validation error check be enough in this case?

src/templates/interfaces/template.ts
CarolinaMoraes: Currently, at the API level, the [schema is using](https://github.com/resend/resend-api/blob/265aadad6546f8f5f1da8aa6183ccd2bb4cc776e/src/routes/templates/validations/schemas.ts#L90-L129) `nullish` and `.optional().nullable()` (both accept `undefined`) for these fields
To me, it makes sense; otherwise, any POST request would have to set each field as null explicitly instead of just not sending it

What do you think?

src/templates/interfaces/template.ts
CarolinaMoraes: Yeah, you're correct! I'll adjust it

src/templates/templates.spec.ts
CarolinaMoraes: For the SDK we don't return it: https://github.com/resend/resend-node/blob/aefde96a54b4e63dd26d095b06ed6103dbfa457b/src/interfaces.ts#L20-L25

src/templates/templates.spec.ts
CarolinaMoraes: I agree, I think it does make sense to reduce these validations -- since we're not testing the API

src/templates/interfaces/template.ts
joaopcm: yes, you're totally right! This works perfectly with the POST endpoint.

The problem is that we're defining this at the template interface level. I imagine that when we start listing templates through the SDK, we want to use this same file/interface, right? If so, this is wrong, since these fields won't be returned as `undefined`, but rather `null` if no value is set in the template entity.

src/templates/templates.spec.ts
CarolinaMoraes: Addressed it here: https://github.com/resend/resend-node/pull/569/commits/03e640ef36fb8f7e20ccfde2b9fc820cf660a5a5

src/templates/interfaces/template.ts
CarolinaMoraes: Makes total sense, I would definitely need to change this once I started to work on another endpoint method 🫠 
I'm glad you caught this sooner

Moved the optional modifier to the create-template-options.interface.ts and now the template interface itself only accepts `null` or truthy values
https://github.com/resend/resend-node/pull/569/commits/3b8a8073c67aeb736a9e1f815985e8b28f9d074b

src/templates/interfaces/template.ts
```diff
@@ -1,22 +1,22 @@
...
-  alias?: string;
-  from?: string;
-  reply_to?: string | string[];
+  text: string | null;
+  status: 'draft' | 'published';
+  variables: TemplateVariable[] | null;
+  alias: string | null;
+  from: string | null;
+  reply_to: string | string[] | null;
   published_at?: string;
```
joaopcm: I think `published_at` might be either string of null, right?

CarolinaMoraes: I absolutely missed it 🤦‍♀️ , thanks for catching it
Changed it here: https://github.com/resend/resend-node/pull/569/commits/102a9692e0844b208929af7f94bda404ea7d8763

src/templates/interfaces/template.ts
```diff
@@ -0,0 +1,23 @@
...
+  id: string;
+  name: string;
+  subject: string | null;
+  html: string;
+  text: string | null;
+  status: 'draft' | 'published';
+  variables: TemplateVariable[] | null;
+  alias: string | null;
+  from: string | null;
+  reply_to: string | string[] | null;
```
joaopcm: `reply_to` is always an array of strings, isn't it?

CarolinaMoraes: Based on how we validate it, it can be either an [array or a single email](https://github.com/resend/resend-api/blob/20d299991637c9bbbf85acdd5c8c27336aee629b/src/routes/emails/validations/email-schema.ts#L381)

joaopcm: But there's a transform that always ensures it is going to be an array [here](https://github.com/resend/resend-api/blob/20d299991637c9bbbf85acdd5c8c27336aee629b/src/routes/emails/validations/email-schema.ts#L382), right?

joaopcm: Also, if you take a look at the database schema, we're always ensuring that this column is an array of strings

CarolinaMoraes: That's correct 🤡 
Fixed it here: https://github.com/resend/resend-node/pull/569/commits/88de606e9754353ee1e51588fa33ba2637a34ad1

PR #571, opened 2025-08-20
feat: add remove templates method

CarolinaMoraes:
## Description 
This PR implements the method to remove templates. When testing, remember that either the ID or an alias can be passed

### DIFF
--- a/src/templates/templates.spec.ts
+++ b/src/templates/templates.spec.ts
@@ -274,29 +274,28 @@ describe('Templates', () => {
         'Failed to render React component. Make sure to install `@react-email/render`',
       );
     });
+  });
 
-    describe('remove', () => {
-      it('removes a template', async () => {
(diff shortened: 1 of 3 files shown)


src/templates/templates.spec.ts
```diff
@@ -274,5 +274,39 @@ describe('Templates', () => {
         'Failed to render React component. Make sure to install `@react-email/render`',
       );
     });
+
+    describe('remove', () => {
```
joaopcm: can we add a test covering an error being returned from the API?

CarolinaMoraes: Added! https://github.com/resend/resend-node/pull/571/commits/a5e97b85dd91abe2be00928eaa1065d7d351e0c4

joaopcm (approved):

PR #574, opened 2025-08-20
feat: add publish method

CarolinaMoraes:
## Summary by cubic
Adds a publish method to Templates to publish a template by ID and return a typed result. Supports Product-403.

- New Features
  - PublishTemplateResponse types for success/error.
  - POST /templates/{id}/publish wired in Templates.publish().
  - Tests for 200 success and 404 not_found cases.

PR #575, opened 2025-08-20
refactor: add chainable operations

CarolinaMoraes:
## Description

In order to achieve something like was [described here](https://github.com/resend/resend-node/pull/574#pullrequestreview-3137421322):

> In order to make the experience of using our SDK smoother, what do you think of implementing additional methods for `create`, `update`, and `duplicate`? I was thinking of something like this:

> ### Create
> ```ts
> await resend.templates.create(/** ... */).publish()
> ```
> 
> ### Update
> ```ts
> await resend.templates.update(/** ... */).publish()
> ```
> 
> ### Duplicate
> ```ts
> await resend.templates.duplicate(/** ... */).publish()
> ```
> 
> Instead of having to call the SDK twice for these operations, like:
> ```ts
> await resend.templates.create(/** ... */);
> await resend.templates.publish(/** ... */);
> ```
> 
> What do you think?

A new class `ChainableTemplateResult` was introduced. It uses PromiseLike to support the native Promise behavior while also enhancing the behavior of the Promise by adding the `publish` method to it.
The method `publish` can now be chained after the `create` and `duplicate` methods are executed. 
(Although the chaining is possible is not required, users that only use `create()` or `duplicate()` won't be affected)

### DIFF
--- a/src/templates/templates.ts
+++ b/src/templates/templates.ts
@@ -1,4 +1,5 @@
 import type { Resend } from '../resend';
+import { ChainableTemplateResult } from './chainable-template-result';
 import type {
   CreateTemplateOptions,
   CreateTemplateResponse,
@@ -25,7 +26,17 @@ export class Templates {
   private renderAsync?: (component: React.ReactElement) => Promise<string>;
   constructor(private readonly resend: Resend) {}
 
-  async create(
+  create(
+    payload: CreateTemplateOptions,
+  ): ChainableTemplateResult<CreateTemplateResponse> {
+    const createPromise = this.performCreate(payload);
+    return new ChainableTemplateResult(createPromise, this.publish.bind(this));
+  }
+  // This creation process is being done separately from the public create so that
+  // the user can chain the publish operation after the create operation. Otherwise, due
+  // to the async nature of the renderAsync, the return type would be
+  // Promise<ChainableTemplateResult<CreateTemplateResponse>> which wouldn't be chainable.
+  private async performCreate(
     payload: CreateTemplateOptions,
   ): Promise<CreateTemplateResponse> {
     if (payload.react) {
@@ -45,11 +56,10 @@ export class Templates {
       );
     }
 
-    const data = await this.resend.post<CreateTemplateResponseSuccess>(
+    return this.resend.post<CreateTemplateResponseSuccess>(
       '/templates',
       payload,
     );
-    return data;
   }
 
   async remove(identifier: string): Promise<RemoveTemplateResponse> {
@@ -66,11 +76,16 @@ export class Templates {
     return data;
   }
 
-  async duplicate(identifier: string): Promise<DuplicateTemplateResponse> {
-    const data = await this.resend.post<DuplicateTemplateResponseSuccess>(
+  duplicate(
+    identifier: string,
+  ): ChainableTemplateResult<DuplicateTemplateResponse> {
+    const promiseDuplicate = this.resend.post<DuplicateTemplateResponseSuccess>(
       `/templates/${identifier}/duplicate`,
     );
-    return data;
+    return new ChainableTemplateResult(
+      promiseDuplicate,
+      this.publish.bind(this),
+    );
   }
 
   async publish(identifier: string): Promise<PublishTemplateResponse> {
--- a/src/templates/chainable-template-result.ts
+++ b/src/templates/chainable-template-result.ts
@@ -1,0 +1,39 @@
+import type { CreateTemplateResponse } from './interfaces/create-template-options.interface';
+import type { DuplicateTemplateResponse } from './interfaces/duplicate-template.interface';
+import type { PublishTemplateResponse } from './interfaces/publish-template.interface';
+
+export class ChainableTemplateResult<
+  T extends CreateTemplateResponse | DuplicateTemplateResponse,
+> implements PromiseLike<T>
+{
+  constructor(
+    private readonly promise: Promise<T>,
+    private readonly publishFn: (
+      id: string,
+    ) => Promise<PublishTemplateResponse>,
+  ) {}
+
+  // If user calls `then` or only awaits for the result of create() or duplicate(), the behavior should be
+  // exactly as if they called create() or duplicate() directly. This will act as a normal promise
+
+  // biome-ignore lint/suspicious/noThenProperty: This class intentionally implements PromiseLike
+  then<TResult1 = T, TResult2 = never>(
+    onfulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | null,
+    onrejected?: ((reason: unknown) => TResult2 | PromiseLike<TResult2>) | null,
+  ): PromiseLike<TResult1 | TResult2> {
+    return this.promise.then(onfulfilled, onrejected);
+  }
+
+  async publish(): Promise<PublishTemplateResponse | null> {
+    const { data, error } = await this.promise;
+
+    if (error) {
+      return {
+        data: null,
+        error,
+      };
+    }
+    const publishResult = await this.publishFn(data.id);
+    return publishResult;
+  }
+}
--- a/src/templates/templates.spec.ts
+++ b/src/templates/templates.spec.ts
@@ -524,11 +524,11 @@ describe('publish', () => {
 
     await expect(resend.templates.publish(id)).resolves.toMatchInlineSnapshot(`
 {
-"data": {
-  "id": "5262504e-8ed7-4fac-bd16-0d4be94bc9f2",
-  "object": "template",
-},
-"error": null,
+  "data": {
+    "id": "5262504e-8ed7-4fac-bd16-0d4be94bc9f2",
+    "object": "template",
+  },
+  "error": null,
 }
 `);
   });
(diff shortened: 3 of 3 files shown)


src/templates/templates.ts
```diff
@@ -45,11 +56,10 @@ export class Templates {
       );
     }
 
-    const data = await this.resend.post<CreateTemplateResponseSuccess>(
+    return this.resend.post<CreateTemplateResponseSuccess>(
       '/templates',
       payload,
     );
-    return data;
   }
```
emiliosheinz, line 63: Not sure if this is required for it to work, but I would rather avoid returning a Promise since, without awaiting, if an error occurs, then the caller function won't appear in the stacktrace. See https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/errorhandling/returningpromises.md for more details

joaopcm: I think that if we change it to `return await ...` it does the job and keep the stacktrace safe, doesn't it?

emiliosheinz: I think so!

lucasfcosta: ++ to this! If anyone wants more context, [this](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/errorhandling/returningpromises.md#advanced-explanation) is helpful.

src/templates/templates.ts
```diff
@@ -25,7 +26,17 @@ export class Templates {
   private renderAsync?: (component: React.ReactElement) => Promise<string>;
   constructor(private readonly resend: Resend) {}
 
-  async create(
+  create(
```
lucasfcosta, line 29: One issue that we might have with these new `PromiseLikes` is that we can miss stack traces. This is related to @emiliosheinz's comment.

I _don't_ think this is a must but I thought it was worth commenting given other people were talking about traces.

In here, we're creating a `PromiseLike` so we're not marking the `create` function as `async` nor awaiting on the actual creation and thus Node won't have its frame on the stack.

You can run the following test to confirm this behaviour:

```ts
it('test async stack trace for create', async () => {
  const mockReactComponent = {
    type: 'div',
    props: { children: 'Welcome!' },
  } as React.ReactElement;

  // Mock renderAsync to throw an error
  mockRenderAsync.mockRejectedValueOnce(new Error('React rendering failed'));

  const payload: CreateTemplateOptions = {
    name: 'Welcome Email',
    react: mockReactComponent,
  };

  const resend = new Resend('re_zKa4RCko_Lhm9ost2YjNCctnPjbLw8Nop');

  // This function will call create and we want to see it in the trace
  const createTemplate = async () => {
    return await resend.templates.create(payload);
  };

  // The error should be thrown and the stack trace should include the createTemplate function, but it doesn't
  try {
    await createTemplate();
    fail('Should have thrown an error');
  } catch (error) {
    // I recommend keeping this log so you can actually copy it and compare
    const stack = (error as Error).stack;
    console.log('Stack trace:', stack);

    expect(error).toBeInstanceOf(Error);
    expect((error as Error).message).toContain('React rendering failed');

    // This fails because `createTemplate` is not part of the stack at all but it should have been
    expect(stack).toContain('createTemplate');
  }
});
```

If you do that you will see that the top of your stack trace has the line for `React rendering failed` (that's what `L63` corresponds to for me, but it does not include `Templates.create` or even `createTemplate (`L74` on the trace.

```
at /Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:63:45
at Generator.next (<anonymous>)
at /Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:8:71
at new Promise (<anonymous>)
at Object.<anonymous>.__awaiter (/Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:4:12)
at Object.<anonymous> (/Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:56:56)
at Promise.then.completed (/Users/lucasfcosta/repositories/resend-node/node_modules/.pnpm/jest-circus@29.7.0/node_modules/jest-circus/build/utils.js:298:28)
<further internals omitted>
```

However, if you use something like [`Error.captureStackTrace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/captureStackTrace) when creating the object you will be able to capture the stack at that point in time and you can then enrich again when throwing. That way, you will see the actual `Templates.create` at the top of the stack, as well as its caller.

```
at Templates.create (/Users/lucasfcosta/repositories/resend-node/src/templates/templates.ts:33:12)
at /Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:74:39
at Generator.next (<anonymous>)
at /Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:8:71
at new Promise (<anonymous>)
at Object.<anonymous>.__awaiter (/Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:4:12)
at createTemplate (/Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:73:41)
at /Users/lucasfcosta/repositories/resend-node/src/templates/templates.spec.ts:79:15
<further internals omitted>
```

If we were to capture the trace what we'd have to do is something like:

```ts
export class ChainableTemplateResult<
  T extends CreateTemplateResponse | DuplicateTemplateResponse,
> implements PromiseLike<T>
{
  private readonly stackTraceTarget: any = {};

  constructor(
    private readonly promise: Promise<T>,
    private readonly publishFn: (
      id: string,
    ) => Promise<PublishTemplateResponse>,
  ) {
    // Capture stack trace excluding this constructor for better error reporting
    if ('captureStackTrace' in Error) {
      Error.captureStackTrace(this.stackTraceTarget, ChainableTemplateResult);
    }
  }
  
  // ...
  
    then<TResult1 = T, TResult2 = never>(
    onfulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | null,
    onrejected?: ((reason: unknown) => TResult2 | PromiseLike<TResult2>) | null,
  ): PromiseLike<TResult1 | TResult2> {
    return this.promise.then(
      onfulfilled,
      (reason: unknown) => {
        // add some logic to recover the trace, message, and name from the captured trace
      }
    );
  }
  
    // ...
}
```

lucasfcosta (approved):
Given we were already talking about traces, I have left a comment about how we could better preserve stack traces when using these `PromiseLike` objects.

Still, I would _not_ block merging this PR because of that given that it introduces some significant complexity for a situation that probably does not happen often and limited benefits.

src/templates/templates.ts
CarolinaMoraes: @emiliosheinz, I didn’t think about it before
Thanks for pointing it out!!

Unfortunately for this approach, it needs to return the promise itself. For the chaining to work, for example `create().publish()`, `create()` must return an object with a publish method immediately

If we use `await`, the `create()` method would be `async`. This method then would return `Promise<ChainableResult>` (and the `ChainableResult` already is a Promise-like object) which we wouldn’t be able to chain since `Promise<ChainableResult>` doesn’t have a publish method, only `ChainableResult` does

I think it's a very valid concern, although the stack trace would still surface that the error came from Templates, it wouldn't have the caller function. The tradeoff is that this is the easier way (in terms of discoverability) to show that a template can be published

For DX, a possible path, would be to use a builder pattern -- not necessarily this eliminates the promise issue, but the issue could be isolated and wouldn't affect anyone using the `create()` directly, only who choose to use something like `.chain().create().publish()`. But this also introduces a new usage pattern into the codebase that we didn't have before, so it still requires some thought

As this is still in preview, and this implementation can still be changed a lot until it actually goes out to general usage, I'll merge this. But I took note of what you said, and I'll make sure to get more feedback about it

CarolinaMoraes: @emiliosheinz, I didn’t think about it before
Thanks a lot for pointing it out

Unfortunately, for this solution, it needs to return the promise itself 
For the chaining to work, for example `create().publish()`, `create()` must return an object with a publish method immediately

If we use await, the `create()` method would be async. This method then would return `Promise<ChainableResult>` (and `ChainableResult` already is a Promise-like object), which we wouldn’t be able to chain since `Promise<ChainableResult>` doesn’t have a publish method, only `ChainableResult` does

As this is still under the preview branch and the implementation can change a lot until it reaches the general public, I’ll merge this
But I took note of what you and @lucasfcosta pointed out and will make sure to gather more feedback about it (maybe we even change the approach of this automatic publishing to something other than chaining)

src/templates/templates.ts
CarolinaMoraes: Thanks a lot @lucasfcosta for the thorough feedback on this (and also the potential solution to this issue)!!
As this branch is still for internal/beta use and won't be available for the general public, this implementation can still change a lot. So I'll go ahead and merge it
But this review didn't go unnoticed. I'll gather more feedback on this method usage to see whether we want to use this Promise-like approach still, or if we should minimize the complexity doing this another way

lucasfcosta: Sounds good @CarolinaMoraes! Thanks for the work on this, I especially liked the elegant typing that went into this 🖤

src/templates/templates.ts
emiliosheinz: Noted! Thank you for the explanation and for looking into it. 

> maybe we even change the approach of this automatic publishing to something other than chaining

IMO, simplifying it by implementing the separate `publish` method that receives and `id` as you did initially, would probably make it more predictable, easier to use and maintain :)

PR #576, opened 2025-08-20
feat: add update template method

CarolinaMoraes:
Besides the method addition itself, this PR also adds two minor changes:
- Fix the expected `fallback_value` type during Template creation
- Moves the `get` template testing block inside the `Templates` block
    

---

## Summary by cubic
Add Templates.update() to patch existing templates and return the updated template id. Also fixes the variable fallback_value type and reorganizes tests under the Templates block.

- New Features
  - Added templates.update(id, payload) using PATCH /templates/:id.
  - Introduced UpdateTemplateOptions/Response types; supports name, subject, html, text, from, reply_to, alias, and variables (key, type, optional fallback_value).
  - Added tests for minimal update, full update, and not-found error.

### DIFF
--- a/src/templates/interfaces/create-template-options.interface.ts
+++ b/src/templates/interfaces/create-template-options.interface.ts
@@ -8,11 +8,18 @@ type TemplateContentCreationOptions = RequireAtLeastOne<{
   react: React.ReactNode;
 }>;
 
+type TemplateVariableCreationOptions = Pick<
+  TemplateVariable,
+  'key' | 'type'
+> & {
+  fallback_value?: string | number | boolean | null;
+};
+
 type TemplateOptionalFieldsForCreation = Partial<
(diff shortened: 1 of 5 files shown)


src/common/interfaces/list-option.interface.ts
```diff
@@ -2,3 +2,13 @@ export interface List<T> {
...
+
+export interface PaginatedList<T> extends List<T> {
+  has_more: boolean;
+}
+
+export interface PaginationOptions {
+  before?: string | null;
+  after?: string | null;
+  limit?: number | null;
+}
```
isabellaaquino: I believe we need to rebase this branch, it contains changes from the list method

CarolinaMoraes: Thanks for pointing it out! Fixed it

PR #581, opened 2025-08-21
feat: pkg-pr-new workflow

gabrielmfern:
Adds in a new Preview Release gh workflow that runs the Node SDK through https://pkg.pr.new to publish it in their repository. It runs on an pull requests, but only publishes if it finds changes inside the `src` directory.

When it actually runs and publishes, a new comment is added to the pull request/commit:

CleanShot 2025-08-21 at 15 33 55@2x

This screenshot is from #582, which tests the workflow out.

It can then be installed and will be the same as running from an npm package. It's also helpful for users to install versions from pull requests to avoid them being blocked.

This is already being used in React Email, works amazingly, and this pull request now standardizes between both.

### DIFF
--- a/.github/scripts/check-pinned-dependencies.mjs
+++ b/.github/scripts/check-pinned-dependencies.mjs
@@ -1,4 +1,5 @@
 import fs from 'node:fs';
+
 const pkg = JSON.parse(await fs.promises.readFile('package.json', 'utf8'));
 const errors = [];
 


CarolinaMoraes (approved):

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,38 @@
+
+name: Preview Release
+on: 
+  push:
+    branches: 
+      - canary
```
CarolinaMoraes, line 6: Can we add a prefix matcher for preview branches? I just saw your [comment here](https://www.notion.so/How-to-release-Node-libraries-700b96705ccd4c9585eefd9c58f99dad?d=246c40d6c4ef8003b54c001cd00e6852&source=copy_link#1ddc40d6c4ef80a58920dbac09d19de9) and thought this same workflow can be used for features that are still in beta/being internally tested (e.g. preview-templates)



```suggestion
  push:
    branches: 
      - canary
      - preview-*
```

gabrielmfern: let's do it!

PR #592, opened 2025-08-22
feat: export new Response type

gabrielmfern:
## Summary by cubic
Expose the Response type from the package entrypoint so consumers can import it directly alongside ErrorResponse. This improves DX and lets apps type responses without deep imports.


xentobias:
My PR would already export the type, but it has been open for quite some time, so I doubt that the repository is actually still actively maintained.

https://github.com/resend/resend-node/pull/547

PR #596, opened 2025-08-22
feat: update send email method to also support templates

CarolinaMoraes:
## Summary by cubic
Add template support to emails.send, enabling delivery by template.id with optional template.variables alongside existing react/html/text flows. Supports PRODUCT-426 individual email by allowing single-send templates with optional from and subject overrides.

- **New Features**
  - Added template object with id and variables properties to CreateEmailOptions and API payloads.
  - Made from and subject optional for template emails; can override template defaults.
  - Enforced mutually exclusive modes (content vs. template) and updated parseEmailToApiOptions to map fields correctly. Added tests for success and error cases.

### DIFF
--- a/src/common/utils/parse-email-to-api-options.spec.ts
+++ b/src/common/utils/parse-email-to-api-options.spec.ts
@@ -2,7 +2,7 @@ import type { CreateEmailOptions } from '../../emails/interfaces/create-email-op
 import { parseEmailToApiOptions } from './parse-email-to-api-options';
 
 describe('parseEmailToApiOptions', () => {
-  it('should handle minimal email with only required fields', () => {
+  it('handles minimal email with only required fields', () => {
     const emailPayload: CreateEmailOptions = {
       from: 'joao@resend.com',
       to: 'bu@resend.com',
@@ -20,7 +20,7 @@ describe('parseEmailToApiOptions', () => {
     });
   });
 
-  it('should properly parse camel case to snake case', () => {
+  it('parses camel case to snake case', () => {
     const emailPayload: CreateEmailOptions = {
       from: 'joao@resend.com',
       to: 'bu@resend.com',
@@ -42,7 +42,7 @@ describe('parseEmailToApiOptions', () => {
     });
   });
 
-  it('should handle template email with template id only', () => {
+  it('handles template email with template id only', () => {
     const emailPayload: CreateEmailOptions = {
       template: {
         id: 'welcome-template-123',
@@ -71,7 +71,7 @@ describe('parseEmailToApiOptions', () => {
     });
   });
 
-  it('should handle template email with template id and variables', () => {
+  it('handles template email with template id and variables', () => {
     const emailPayload: CreateEmailOptions = {
       template: {
         id: 'newsletter-template-456',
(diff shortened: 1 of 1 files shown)


src/common/utils/parse-email-to-api-options.ts
```diff
@@ -25,12 +25,15 @@ export function parseEmailToApiOptions(
...
     reply_to: email.replyTo,
     scheduled_at: email.scheduledAt,
     subject: email.subject,
     tags: email.tags,
-    text: email.text,
+    text: 'text' in email ? email.text : undefined,
     to: email.to,
+    template_id: 'template_id' in email ? email.template_id : undefined,
+    template_variables:
+      'template_variables' in email ? email.template_variables : undefined,
```
emiliosheinz: Just out of curiosity, what is the difference between directly using `email.html` and `'html' in email ? email.html : undefined`?

CarolinaMoraes: @emiliosheinz 
Good question, I really would appreciate some feedback on this

So here is the scenario: `CreateEmailOptions` is a union type with two mutually exclusive cases. So if it has template properties, it won't have react, html, and text properties (and vice-versa)

```ts
export type CreateEmailOptions =
  | ((RequireAtLeastOne<EmailRenderOptions> & CreateEmailBaseOptions) & {
      template_id?: never;
      template_variables?: never;
    })
  | ((EmailTemplateOptions & CreateEmailBaseOptionsWithTemplate) & {
      react?: never;
      html?: never;
      text?: never;
    });
```

In the snippet you highlighted, if we didn't have this extra check, if we passed template properties and then we tried to access `email.html` directly, its type would be `never`

But as JS itself doesn’t have types, at runtime, this `html` property would become `undefined` anyway. I just thought this way was more explicit and would prevent mismatch between Typescript and runtime

What do you think?

emiliosheinz: I see, thank you for clarifying. Well, I have no strong preference, but if TypeScript doesn't yell at you if you simply do `html: email.html`, I would probably leave it like that, as the other option is more verbose.

src/common/interfaces/email-api-options.interface.ts
```diff
@@ -22,4 +22,6 @@ export interface EmailApiOptions {
   scheduled_at?: string;
   tags?: Tag[];
   attachments?: EmailApiAttachment[];
+  template_id?: string;
+  template_variables?: Record<string, unknown>;
```
joaopcm: BTW, we've changed the API payload contract to accept an object field `template`, which includes the required `id` field and an optional `Record<string, string | number | boolean>` `variables`

CarolinaMoraes: Nice! I updated the SDK method to send this new shape but it will also accept an object like this as the input
https://github.com/resend/resend-node/pull/596/commits/91cbb741b68dfabf0d160674bee14889d4a800a7

src/common/utils/parse-email-to-api-options.spec.ts
```diff
@@ -41,4 +41,117 @@ describe('parseEmailToApiOptions', () => {
       scheduled_at: 'in 1 min',
     });
   });
+
+  it('should handle template email with template id only', () => {
```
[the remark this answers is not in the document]
CarolinaMoraes: Addressed on https://github.com/resend/resend-node/pull/596/commits/7b4e65768ed334f95cd4cee179f4b53eda160a42

PR #599, opened 2025-08-25
feat: adds permissive mode for batch API

pedro-stramantinoli:
This PR is related to this [feature](https://github.com/resend/resend-api/pull/2025).

Basically it should allow to call batch API with header `x-batch-validation`.
If the header value is `permissive`, it should expect to receive a field `errors` with `data`.
This should *NOT* cause a breaking change.

For that, if no header is provided, we guarantee that no field `errors` will be returned.

### DIFF
--- a/src/batch/batch.ts
+++ b/src/batch/batch.ts
@@ -52,6 +52,15 @@ export class Batch {
       options,
     );
 
+    // If the permissive header is not set, remove errors to maintain backward compatibility
+    if (data.data && options.headers?.['x-batch-validation'] !== 'permissive') {
+      const { errors, ...dataWithoutErrors } = data.data;
+      return {
+        ...data,
+        data: dataWithoutErrors,
+      };
+    }
+
     return data;
   }
 }
--- a/src/test-utils/mock-fetch.ts
+++ b/src/test-utils/mock-fetch.ts
@@ -63,6 +63,20 @@ export function mockSuccessResponse<T>(
 }
 
 /**
+ * Mock successful response with a custom status code
+ */
+export function mockSuccessWithStatusCode<T>(
+  data: T,
+  status: number,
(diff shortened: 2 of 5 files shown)


src/batch/batch.ts
```diff
@@ -52,6 +52,15 @@ export class Batch {
       options,
     );
 
+    // If the permissive header is not set, remove errors to maintain backward compatibility
+    if (data.data && options.headers?.['x-batch-validation'] !== 'permissive') {
```
gabrielmfern: Can we use an option for this instead of a header like this? The idea is to reduce friction for users so that they don't have to look into our documentation to find out about this feature, and so that things play better with typescript.

src/batch/batch.ts
```diff
@@ -52,6 +52,15 @@ export class Batch {
...
+      const { errors: _errors, ...dataWithoutErrors } = data.data;
```
gabrielmfern: See https://github.com/resend/resend-api/pull/2025/files#r2303798877

src/test-utils/mock-fetch.ts
```diff
@@ -62,6 +62,20 @@ export function mockSuccessResponse<T>(
...
+/**
+ * Mock successful response with a custom status code
+ */
+export function mockSuccessWithStatusCode<T>(
+  data: T,
+  status: number,
+  options: MockFetchOptions = {},
+): void {
+  mockFetchWithRateLimit(JSON.stringify(data), {
+    status,
```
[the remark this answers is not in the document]
pedro-stramantinoli: Options does not contain status

gabrielmfern (approved):
LGTM! But since I made so many changes I think it would be nice to get the eyes of somebody else too

PR #608, opened 2025-08-29
fix: changes headers name to keep standard

pedro-stramantinoli:
Add prefix `resend`  to headers name in order to keep our standard
    

---

## Summary by cubic
Renamed the batch validation header to x-resend-batch-validation to align with our x-resend-* standard. Updated request headers, tests, and inline docs; the batchValidation option and behavior are unchanged.

### DIFF
--- a/src/batch/batch.spec.ts
+++ b/src/batch/batch.spec.ts
@@ -319,7 +319,7 @@ describe('Batch', () => {
       const request = lastCall[1];
       expect(request).toBeDefined();
       const headers = new Headers(request?.headers);
-      expect(headers.get('x-batch-validation')).toBe('permissive');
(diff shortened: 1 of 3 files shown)


pedro-stramantinoli:
> Hey Pedro!
> 
> Don't we need to change this line as well so the API can pick it up?
> 
> https://github.com/resend/resend-api/blob/424ee12e5abea86a78a06fa4b0dbee7730e9ef18/src/routes/emails/batch/create.ts#L446-L449

Oh yes, it's being changed in this [PR](https://github.com/resend/resend-api/pull/2047)
Also, this header is not being used yet

lucasfcosta (approved):
Nice! Thanks for the response 🖤

PR #611, opened 2025-08-31
feat: add async iterator support to emails.list method

vieiralucas:
Add PaginatedRequest class with async iteration capabilities to handle paginated email listing. Includes automatic retry logic for rate limits, forward and backward pagination support.
    

---
## Summary by cubic
Adds async iterator support to `emails.list` via a new `PaginatedRequest` that acts as both a Promise and an AsyncIterable. Enables simple forward/backward pagination with automatic rate-limit retries and exposes rate-limit metadata on all SDK responses. Aligns with ENG-4362.

- **New Features**
  - `emails.list` now returns `PaginatedRequest<ListEmail>` (Promise + AsyncIterable).
  - Iterate pages with for‑await‑of; `after`/`before` cursors set from item IDs automatically.
  - Retries on `rate_limit_exceeded` using `Retry-After` when present; `retryAfter` surfaced on errors.
  - All responses include `rateLimiting` metadata when headers are available.

- **Migration**
  - `await emails.list(...)` still returns the first page.
  - The return type is `PaginatedRequest`; update code that assumed a plain Promise.
  - To stream all pages: `for await (const page of emails.list(opts)) { ... }`

<sup>Written for commit dcbaf0a82dd49d4b2a6be80276954d5e18111a30. Summary will update on new commits.</sup>

### DIFF
--- a/integrations/nextjs/package.json
+++ b/integrations/nextjs/package.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "next": "16.1.6",
+    "next": "16.1.7",
     "react": "19.2.4",
     "react-dom": "19.2.4",
     "resend": "../.."
--- a/package.json
+++ b/package.json
@@ -74,13 +74,13 @@
     "@types/react": "19.2.14",
     "cross-env": "10.1.0",
     "dotenv": "17.3.1",
-    "pkg-pr-new": "0.0.65",
+    "pkg-pr-new": "0.0.66",
     "react": "19.2.4",
     "react-dom": "19.2.4",
     "rimraf": "6.1.3",
     "tsdown": "0.21.0",
     "typescript": "5.9.3",
-    "vitest": "4.0.18",
+    "vitest": "4.1.0",
     "vitest-fetch-mock": "0.4.5"
   },
   "packageManager": "pnpm@10.30.3"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -46,8 +46,8 @@ importers:
         specifier: 17.3.1
         version: 17.3.1
       pkg-pr-new:
-        specifier: 0.0.65
-        version: 0.0.65
+        specifier: 0.0.66
+        version: 0.0.66
       react:
         specifier: 19.2.4
         version: 19.2.4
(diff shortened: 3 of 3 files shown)


src/common/paginated-request.ts
```diff
@@ -0,0 +1,166 @@
...
+        // if first request was rate limited, wait and continue
+        // we'll enter the pagination loop without advancing the cursor
+        const retryAfter = initialPage.error.retryAfter * 1000;
+        await new Promise((resolve) => {
+          setTimeout(resolve, retryAfter);
+        });
+      } else {
+        // TODO: how do we expose this error to the caller?
+        console.error('Error fetching initial page:', initialPage.error);
+        return;
```
vieiralucas: My original idea was to make an iterator on `T`, but since we return errors as value we can't really have that without swallowing the errors.

src/common/paginated-request.ts
```diff
@@ -0,0 +1,166 @@
...
+    // Start iterating through subsequent pages
+    let currentOptions = { ...this.initialOptions };
+    let hasMore = true;
+
+    while (hasMore) {
+      const response = await this.fetchPageWithRetry(currentOptions);
+      if (response.error) {
+        // TODO: how do we expose this error to the caller?
+        console.error('Error fetching page:', response.error);
+        break;
```
vieiralucas: So this is another instance of the same problem as above

vieiralucas:
@gabrielmfern about handling errors:

Unless we want to use exceptions I though of 2 alternatives for the api to expose the error as value:

1. iterate on page requests
```typescript
for await (const page of resend.emails.list().iterate()) {
  if (page.error) {
    // handle the error
    break;
  }

  // iterate page items  
  for (const item of page.data) {
  }
}
```

2. iterate on items, but yield a `Response<T>`
```typescript
for await (const item of resend.email.list().iterate()) {
  if (item.error) {
    // handle the error
    break;
  }
  
  const email = item.data;
}
```

emiliosheinz:
@vieiralucas From an outside perspective, I’d love it if you could explain in more detail what problems we’re trying to solve with this change, and why our current implementation of the emails listing API doesn’t address them. If it is just for an easier developer experience, I think that is a totally valid reason. 

Also, if we agree on the solution, I’d lean toward iterating on the pages (option 1) rather than on the items. The reason is that, technically, the implementation isn’t fetching one item at a time. It’s fetching new pages as needed. Because of that, showing an error tied to a specific item feels a bit misleading, since the error actually occurs when fetching the entire next page, not the individual item.

Does that make sense?

vieiralucas:
Hi @emiliosheinz, thanks for the comment.

This is basically us wanting to provide better DX, just pushed a new commit that yields the page instead of the item btw, but we're still doing some discussions internally about this PR.

PR #617, opened 2025-09-05
fix: typo in `invalid_api_key` error

gabrielmfern:
Before this we had `invalid_api_Key` in the type, while the [docs](https://resend.com/docs/api-reference/errors#invalid-api-key) say it should be `invalid_api_key`. It also seems like we don't return this any more, and just return `validation_error` when the API Key is invalid/expired, but fixing the type anyways for now.


gabrielmfern:
@vieiralucas Nope, the API never returned `invalid-api-Key`, and we just return the value from the API – it should be a patch change.

That map of status codes is actually never used right now, so it's basically just types

PR #618, opened 2025-09-05
chore: revert `rateLimiting` API to avoid blocking release channel

gabrielmfern:
It reverts most changes that were made in #562, but it doesn't revert the response mocking utilities that it introduced. This might interfere with the implementation in #611.


src/batch/interfaces/create-batch-options.interface.ts
```diff
@@ -40,4 +40,11 @@ export type CreateBatchSuccessResponse<
...
 export type CreateBatchResponse<Options extends CreateBatchRequestOptions> =
-  Response<CreateBatchSuccessResponse<Options>>;
+  | {
+      data: CreateBatchSuccessResponse<Options>;
+      error: null;
+    }
+  | {
+      data: null;
+      error: ErrorResponse;
+    };
```
pedro-stramantinoli, line 50: If I understood this correctly, it means we cannot receive a response with `data` and `error` not `null`?
If so, than I think this is not correct, because if the `x-batch-validation` header is `permissive` we can have `data` being empty array or normal array and `error` can be empty array or normal array.
So no scenario `data` would be `null`.

gabrielmfern: The SDK actually wraps the response that comes from our API in the `data` property, so it should be fine, this pull request doesn't change it except for removing the `rateLimiting` property. So, this type is actually the same type as it was before, but inlined, like it was before #562

pedro-stramantinoli (approved):

Issue #625, opened 2025-09-17
Turbopack resolves optional `@react-email/render` import in v6.1.0, causing “Module not found”


mmstroik:
### Describe the Bug

In `resend@6.1.0`, I think moving the dynamic import of `@react-email/render` into a shared `src/render.ts` helper (introduced in [PR #620](https://github.com/resend/resend-node/pull/620)) causes Next.js 15 Turbopack to resolve the literal specifier inside that helper module. This triggers a module resolution error even when e.g. users do not pass a `react` payload (using only `html` and/or `text`). The same usage worked in `6.0.3`.

Tbf the onus here might be on vercel to fix the eager compiling with turbopack (this isn't an issue for non-turbopack builds). And there is a fairly simple workaround by just adding "render" to the serverExternalPackages array in the next.config file or just lazy-loading resend.

### Link to the code that reproduces this issue

https://github.com/mmstroik/resend-turbopack-repro

### To Reproduce

1. create a Next.js 15 app (App Router) with npx create-next-app (or clone the linked reproduction repo and skip to step 3)
2. install `resend@6.1.0` and add an API route (route.ts file) with:
```route.ts
import { Resend } from "resend";
new Resend();
```
3. Run `next build --turbopack` (or run dev server with --turbopack and call the api route)
4. Observe error

### Expected Behavior

The dynamic import should not be resolved during build and the route should succeed without installing `@react-email/render`.

### What's your node version? (if relevant)

22.16.5

krisnaw:
Having the same issue, did you find temporary solution @mmstroik ?

emiliosheinz:
The easiest temporary solution — although not ideal — is to install `@react-email/render`. As a side note, since I’ve seen many people expressing confusion about this dependency, I’ve opened a discussion proposing an alternative approach that, in my opinion, would make things much clearer and easier to manage. Feel free to share your thoughts there!

Discussion: https://github.com/resend/resend-node/discussions/588

gabrielmfern:
Can indeed reproduce that this is a regression from `6.0.3` to `6.1.0` introduced in #620

gabrielmfern:
Should be fixed in [`resend@6.2.2`](https://github.com/resend/resend-node/releases/tag/v6.2.2)!

PR #627, opened 2025-09-22
fix(render): handle optional peer dependency @react-email/render safely

paaauldev:
Prevent crash when @react-email/render is not installed by checking module existence before calling `render`. This ensures compatibility with optional peer dependencies.

Refs: resend/react-email#2474, resend/resend-node#625
    

---

## Summary by cubic
Handle optional peer @react-email/render safely to prevent crashes when the package isn’t installed. We now validate the dynamic import and the render function, and return a clear error instead of throwing, keeping projects without this dependency working.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,5 +61,5 @@
     "vitest": "3.2.4",
     "vitest-fetch-mock": "0.4.5"
   },
-  "packageManager": "pnpm@10.16.1"
+  "packageManager": "pnpm@10.17.0"
 }


gabrielmfern:
Did you also try it out on the reproductions for resend/react-email#2474 and resend/resend-node#625 to make sure it works? You can use pkg.pr.new's preview version for that:

```zig
npm i https://pkg.pr.new/resend/resend-node/resend@627
```

gabrielmfern:
Accidentally closed, sorry

paaauldev:
Just fixed with last commit please review @gabrielmfern

src/render.ts
```diff
@@ -1,26 +1,34 @@
...
-          reject(
-            Error(
-              'Failed to render React component. Make sure to install `@react-email/render`',
-            ),
-          );
-          return;
-        }
+    // Create an indirect require that bundlers won't statically analyze
+    // eslint-disable-next-line @typescript-eslint/no-implied-eval
+    const req = Function('m', 'return require(m);') as (m: string) => any;
```
gabrielmfern, line 5: Why use a require for this?

src/render.ts
```diff
@@ -1,26 +1,34 @@
...
+        return;
+      }
+    } catch {
+      // ignore and fallback below
+    }
+
+    // Fallback: use react-dom/server to render to static markup
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const { renderToStaticMarkup } = req('react-dom/server');
```
gabrielmfern, line 32: I don't particularly like falling back to rendering with `react-dom` directly. React Email does a lot of work to actually use `react-dom/server` on all environments without any issues, if we do this here we're going to have to maintain a duplicate of the code that's already there.

src/render.ts
paaauldev: Uses indirect require to handle optional dependencies safely: tries @react-email/render if available, falls back to react-dom/server if not, and checks module existence at runtime with try/catch. This avoids static bundling issues, prevents build/runtime errors, and works in both CJS and ESM environments.

If you want, I can also make an even snappier one-line version for the PR.

src/render.ts
paaauldev: Ok, I removed the direct react-dom/server fallback and rely on @react-email/render as the single rendering path. React Email already handles cross-environment quirks and we avoid duplicating that logic here. If @react-email/render isn’t installed, we now fail fast with a clear message and recommend installing it via: pnpm add @react-email/render@latest

src/render.ts
gabrielmfern: `import` does work in CJS, though, do you have an example where it doesn't to include this?

Also, I don't think `require` always works on ESM either.

src/render.ts
gabrielmfern: Maybe you forgot to commit? The pull request doesn't seem to have been updated with this

gabrielmfern:
Closing this due to the inactivity, since I don't think the author wants this change anymore.

PR #632, opened 2025-09-27
feat(ci): low-hanging fruit for performance

lucasfcosta:
This adds a few quick wins to try make CI take up less resources and not be clogged as requested by @gabrielmfern. It:

* Uses smaller buildjet instances for small tasks
* Cancels workflows when pushing to a branch with previous executions (useful to avoid useless runs in PRs)
* Uses a smaller node image

Note that these are just simple quick-wins, what will really help with CI is changing the runners, which we'll do separately.

Also note that release workflows are mostly unchanged to avoid canceling releases in the middle or using smaller machines for building.
    

---

## Summary by cubic
Reduce CI load and queue time by canceling superseded runs and right-sizing runners and images.

- **Refactors**
  - Add concurrency groups (per workflow/ref) to cancel in-progress runs.
  - Use buildjet-2vcpu-ubuntu-2204 for lint, pin-dependencies-check, and PR title check; tests stay on 4 vCPU.
  - Switch container image to node:22-slim across workflows.


gabrielmfern (approved):
Thank you!

PR #641, opened 2025-09-30
feat: get inbound emails v0

lucasfcosta:


src/attachments/attachments.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+        fetchMock.mockOnce(JSON.stringify(response), {
+          status: 404,
+          headers: {
+            'content-type': 'application/json',
+            Authorization: 'Bearer re_zKa4RCko_Lhm9ost2YjNCctnPjbLw8Nop',
+          },
+        });
+
+        const result = await resend.attachments.get({
+          emailId: '61cda979-919d-4b9d-9638-c148b93ff410',
```
zenorocha: In the spirit of keeping everything isolated during the alpha testing, and not causing confusion with outbound, how about we rename `emailId` to `inboundId`?

vcapretz (approved):

PR #664, opened 2025-10-08
chore: improve PR title check error

TyMick:
The example in the error message isn't actually compliant with the rules. Most confusingly to me, the title portion after the colon started with a capital letter, while the rules require it to start with a lowercase letter. But while I'm at it, I figured we may as well use `feat` as the example's type instead of `type`, since `type` isn't an allowed type.


.github/scripts/pr-title-check.js
```diff
@@ -11,7 +11,7 @@ const validateTitle = (title) => {
   if (!isValidType(title)) {
     console.error(
       `PR title does not follow the required format.
-      example: "type: My PR Title"
+      example: "feat: my PR title"
```
gabrielmfern: how about it we change to saying `[...] the required format [type]: [title]. Example: "fix: email compatibility issue"`?

TyMick: Even better! 🙌🏼

TyMick: How does https://github.com/resend/resend-node/pull/664/commits/7120123d301ae19cf03b87aa41d3ea7ab44df7c9 look, for further disambiguation?

gabrielmfern: LGTM!

PR #666, opened 2025-10-08
feat: move emails inbound method to emails.receiving.<method>

lucasfcosta:
This moves the `resend.inbound` methods so that they're nested under `resend.emails.receiving.<method>` as per the RFC. 

It also makes it so that we'll use the `/emails/receiving` route rather than the old `/inbound` that we aliased on the API.

Also, worth mentioning that I did change the response to be `snake_cased` like all the other methods on the SDK. I previously was converting to camel cased because we discussed that on Slack but it was actually inconsistent with other methods.
    

---

## Summary by cubic
Moved inbound email methods under resend.emails.receiving and switched API calls to /emails/receiving. This follows the RFC and removes the top-level inbound API.

- **Refactors**
  - Renamed Inbound to Receiving and exposed as emails.receiving.
  - Removed Resend.inbound.
  - Switched GET route to /emails/receiving/:id.
  - Updated tests and interface import paths; exports now point to emails/receiving.

- **Migration**
  - Change resend.inbound.get(id) to resend.emails.receiving.get(id).
  - Update imports from 'inbound/interfaces' to 'emails/receiving/interfaces'.


gabrielmfern:
the evil pull request number

PR #670, opened 2025-10-09
feat: add pagination for inbound email attachments

lucasfcosta:
This PR is the SDK complement to https://github.com/resend/resend-api/pull/2192. It adds pagination to the inbound email attachments.

I also used this opportunity to:

* Update the `object` to `list` for the method that lists attachment (I did that on the API PR too)
* Use the new `/emails/receiving/:emailId/attachments` alias from the API instead of the old `/emails/inbound/:emailId/attachments`
    

---

## Summary by cubic
Added pagination for inbound email attachments in the SDK and switched to the new /emails/receiving endpoints. Aligns with PRODUCT-710 to enable cursor-based navigation and reduce large payloads.

- **New Features**
  - attachments.receiving.list now supports limit, before, and after.
  - Uses /emails/receiving/:emailId and /emails/receiving/:emailId/attachments/:id.
  - Builds query params only when pagination options are provided.

- **Migration**
  - List responses now return object: 'list' and has_more. Update any checks relying on the previous object value.

### DIFF
--- a/src/attachments/receiving/receiving.ts
+++ b/src/attachments/receiving/receiving.ts
@@ -1,3 +1,4 @@
+import { buildPaginationQuery } from '../../common/utils/build-pagination-query';
 import type { Resend } from '../../resend';
 import type {
   GetAttachmentOptions,
@@ -17,7 +18,7 @@ export class Receiving {
     const { emailId, id } = options;
 
     const data = await this.resend.get<GetAttachmentResponseSuccess>(
-      `/emails/inbound/${emailId}/attachments/${id}`,
+      `/emails/receiving/${emailId}/attachments/${id}`,
     );
 
     return data;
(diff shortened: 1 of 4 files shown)


src/attachments/receiving/receiving.ts
```diff
@@ -28,9 +29,12 @@ export class Receiving {
   ): Promise<ListAttachmentsResponse> {
     const { emailId } = options;
 
-    const data = await this.resend.get<ListAttachmentsResponseSuccess>(
-      `/emails/inbound/${emailId}/attachments`,
-    );
+    const queryString = buildPaginationQuery(options);
+    const url = queryString
+      ? `/emails/receiving/${emailId}?${queryString}`
```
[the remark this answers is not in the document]
lucasfcosta: Good catch little friend. Thanks.

src/attachments/receiving/receiving.ts
```diff
@@ -17,7 +18,7 @@ export class Receiving {
     const { emailId, id } = options;
 
     const data = await this.resend.get<GetAttachmentResponseSuccess>(
```
bukinoshita, line 20: Why is it calling an API and not the DB here?

lucasfcosta: Not sure I understand this comment. This is the `get` method that we call on `resend.emails.receiving.get` so it can't connect to a DB? Maybe I'm missing something? 🤔

lucasfcosta: For reference, the existing `get` method for the outbound emails does the same call but to the outbound route:

```
  async get(id: string): Promise<GetEmailResponse> {
    const data = await this.resend.get<GetEmailResponseSuccess>(
      `/emails/${id}`,
    );

    return data;
  }
```

PR #676, opened 2025-10-14
feat: download from API's signed URLs instead of proxied routes

lucasfcosta:
Before https://github.com/resend/resend-api/pull/2199 we were proxying contents directly through our API. For various reasons related to reliability and overall separation of concerns, we started using signed URLs as explained in the aforementioned PR.

This PR makes our SDK automatically download from those URLs and encode content as base64 so it can automatically be used by the `resend.send` method.

I'm also bumping the version here so we don't have to do a separate PR for that.
    

---

## Summary by cubic
Switch SDK attachment downloads to API-provided signed URLs, then return base64 content to preserve the SDK response shape. Aligns with PRODUCT-796 by removing proxied content for better reliability.

- **New Features**
  - get: fetches metadata with download_url, downloads file, returns base64 content.
  - list: downloads each attachment via signed URL and returns base64 content.
  - Types: API responses now include download_url; SDK responses still expose content (base64).
  - Errors: returns application_error if a signed URL download fails.

- **Dependencies**
  - Bump to 6.2.0-canary.4.

### DIFF
--- a/src/attachments/receiving/receiving.ts
+++ b/src/attachments/receiving/receiving.ts
@@ -62,7 +62,11 @@ export class Receiving {
       return apiResponse;
     }
 
-    const { download_url, ...otherFields } = apiResponse.data.data;
+    const {
+      expires_at: _expires_at,
+      download_url,
+      ...otherFields
+    } = apiResponse.data.data;
     const downloadResult = await this.downloadAttachment(download_url);
     if (downloadResult.type === 'error') {
       return {
@@ -106,7 +110,11 @@ export class Receiving {
 
     const attachmentsWithContent = [];
     for (const attachment of apiResponse.data.data) {
-      const { download_url, ...otherFields } = attachment;
+      const {
+        expires_at: _expires_at,
+        download_url,
+        ...otherFields
+      } = attachment;
       const downloadResult = await this.downloadAttachment(download_url);
       if (downloadResult.type === 'error') {
         return {
--- a/src/attachments/receiving/interfaces/attachment.ts
+++ b/src/attachments/receiving/interfaces/attachment.ts
@@ -6,3 +6,13 @@ export interface InboundAttachment {
   content_id?: string;
   content: string; // base64
 }
+
+export interface ApiInboundAttachment {
+  id: string;
+  filename?: string;
+  content_type: string;
+  content_disposition: 'inline' | 'attachment';
+  content_id?: string;
+  download_url: string;
+  expires_at: string;
+}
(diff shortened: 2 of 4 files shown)


src/attachments/receiving/receiving.ts
```diff
@@ -34,8 +66,43 @@ export class Receiving {
...
-    const data = await this.resend.get<ListAttachmentsResponseSuccess>(url);
+    const apiResponse = await this.resend.get<ListAttachmentsApiResponse>(url);
+
+    if ('error' in apiResponse && apiResponse.error) {
+      return apiResponse;
+    }
+
+    const attachmentsWithContent = [];
+    for (const attachment of apiResponse.data.data) {
+      const downloadResponse = await fetch(attachment.download_url);
```
emiliosheinz: Unlike `this.resend.get`, which wraps its internal `fetch` calls with a `try/catch` and normalizes network errors into the standard response format, this direct `fetch` invocation can `throw` under certain edge cases like DNS resolution failures, offline status, or other low-level network issues.

If unhandled, these exceptions would bypass the method’s usual error normalization flow, which may conflict with its original design of consistently returning an `{ data, error }` structure.

I'm not sure if the omission of the `try/catch` blocks was intentional; if not, I would consider wrapping it with them to avoid unexpected errors being thrown.

lucasfcosta: Great point!

src/attachments/receiving/receiving.ts
```diff
@@ -34,8 +102,43 @@ export class Receiving {
...
+    }
+
+    const attachmentsWithContent = [];
+    for (const attachment of apiResponse.data.data) {
+      const {
+        expires_at: _expires_at,
+        download_url,
+        ...otherFields
+      } = attachment;
+      const downloadResult = await this.downloadAttachment(download_url);
```
emiliosheinz, line 118: Not sure if performance is a concern for this, but you might want to put this under a `Promise.all` or `Promise.allSettled` so that all the attachments are downloaded in parallel. 

I missed that during my first review, sorry 😥

lucasfcosta: Yeah, we discussed this internally as I was reviewing with @vieiralucas. For now we're going to download sequentially. The reason being that we want to review whether the SDK should download attachments at all. We're inclined to consider the download part as something that the user controls. That way they can download in whichever way they find best.

For now we decided to do it sequentially because it's a safe default. We're going to discuss this again today in one of our syncs.

PR #683, opened 2025-10-16
feat: new 'suspended_api_key' in error name type

gabrielmfern:
Adds in the new `suspended_api_key` error name to the type union of the possible error names that the API can respond with. 

    

---

## Summary by cubic
Added 'suspended_api_key' to the RESEND_ERROR_CODE_KEY union so suspended API key errors are typed correctly. Removed the unused RESEND_ERROR_CODES_BY_KEY status code map.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -1,21 +1,24 @@
-export type RESEND_ERROR_CODE_KEY =
-  | 'missing_required_field'
-  | 'invalid_idempotency_key'
-  | 'invalid_idempotent_request'
-  | 'concurrent_idempotent_requests'
-  | 'invalid_access'
-  | 'invalid_parameter'
-  | 'invalid_region'
-  | 'rate_limit_exceeded'
-  | 'missing_api_key'
-  | 'invalid_api_key'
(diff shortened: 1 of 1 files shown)


src/interfaces.ts
```diff
@@ -1,23 +1,21 @@
...
-  invalid_from_address: 403,
-  validation_error: 403,
-  not_found: 404,
-  method_not_allowed: 405,
-  application_error: 500,
-  internal_server_error: 500,
-} as const;
-
-export type RESEND_ERROR_CODE_KEY = keyof typeof RESEND_ERROR_CODES_BY_KEY;
+export type RESEND_ERROR_CODE_KEY =
```
bukinoshita: why did we change from `const` to `type`? I believe we used to use type and changed to const

bukinoshita: Maybe in this PR we just add the `suspended_api_key` and open a different one that change if we feel it makes sense?

gabrielmfern: yeah!

bukinoshita (approved):

PR #689, opened 2025-10-17
feat: introduce the /segment API endpoints

Cisneiros:
This PR introduces the methods on `resend.segments.*` to succeed `resend.audiences.*`.

The `resend.audiences.*` methods are kept for backwards compatibility, and marked as deprecated. Test were introduced to make sure both sets of methods continue to work the same as expected.

For the Broadcast API, using either `segment_id` or `audience_id` is supported. At least one is required.
    

---

## Summary by cubic
Introduces the Segments API and SDK (resend.segments.*) to replace audiences, with backwards-compatible deprecations. Broadcasts now accept segmentId (or audienceId) so you can target segments directly.

- **New Features**
  - Segments SDK: create, get, list, remove.
  - Broadcasts: support segmentId; segment_id returned in list/get.
  - Contacts: audiences replaced with segments (add/list/remove).
  - Audiences methods kept and marked deprecated.
  - Added unit and integration tests for segments.

- **Migration**
  - Use resend.segments.* instead of resend.audiences.*.
  - For broadcasts, prefer segmentId; audienceId still works. At least one is required.
  - Access contacts via contacts.segments instead of contacts.audiences.
  - API responses still use object: "audience" for compatibility.

### DIFF
--- a/src/segments/interfaces/remove-segment.interface.ts
+++ b/src/segments/interfaces/remove-segment.interface.ts
@@ -2,7 +2,7 @@ import type { ErrorResponse } from '../../interfaces';
 import type { Segment } from './segment';
 
 export interface RemoveSegmentResponseSuccess extends Pick<Segment, 'id'> {
-  object: 'audience';
+  object: 'segment';
   deleted: boolean;
 }
 
--- a/src/segments/__recordings__/Audiences-Integration-Tests-create-creates-an-audience_3138926239/recording.har
+++ b/src/segments/__recordings__/Audiences-Integration-Tests-create-creates-an-audience_3138926239/recording.har
@@ -25,7 +25,7 @@
             },
             {
               "name": "user-agent",
-              "value": "resend-node:6.3.0-canary.0"
+              "value": "resend-node:6.3.0-canary.1"
             }
           ],
           "headersSize": 181,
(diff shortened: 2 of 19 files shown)


src/resend.ts
```diff
@@ -30,7 +30,11 @@ export class Resend {
 
   readonly apiKeys = new ApiKeys(this);
   readonly attachments = new Attachments(this);
-  readonly audiences = new Audiences(this);
+  readonly segments = new Segments(this);
+  /**
+   * @deprecated Use segments instead
+   */
+  readonly audiences = this.segments;
```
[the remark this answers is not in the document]
Cisneiros: Types were renamed only, they still return the same fields.

src/segments/interfaces/remove-segment.interface.ts
```diff
@@ -0,0 +1,17 @@
+import type { ErrorResponse } from '../../interfaces';
+import type { Segment } from './segment';
+
+export interface RemoveSegmentResponseSuccess extends Pick<Segment, 'id'> {
+  object: 'audience';
```
bukinoshita: Is this correct?

Cisneiros: Updated the object type

bukinoshita (approved):

PR #690, opened 2025-10-17
feat: support object and list variables in email sending methods

joaopcm:
- **feat(email templates): support object and list variable types for template variables**
- **test(batch): add spec for batch emails with object and list template variables**

    

---

## Summary by cubic
Added support for object and list template variables when sending emails and batch. Templates can now use nested objects and arrays in variable values.

- **New Features**
  - Extended EmailApiOptions and CreateEmailOptions to accept nested objects and list variables (TemplateVariableListFallbackType).
  - Added tests for single and batch sends to verify payloads with complex variables.


CarolinaMoraes:
As we're not going to focus on template languages now, I'm closing this pull request. We can reopen it at a better time

PR #693, opened 2025-10-17
feat: add verify webhooks to main

lucasfcosta:
The webhook verification method is useful beyond canary, so I'd love to cherry-pick it to main so we can update the docs for everyone to use this method.

I'm not sure, however, if this is how we usually port small changes from canary to `main` for early releases of `latest`. I have already tested this with a local package and it works fine btw.

Anything I should be aware here or should we just merge this and `npm publish --tag latest`?
    

---

## Summary by cubic
Add a webhook verification helper to the SDK (main). This exposes resend.webhooks.verify to validate Resend webhooks using Svix and bumps the package to 6.2.0.

- **New Features**
  - New Webhooks.verify({ payload, headers: { id, timestamp, signature }, webhookSecret }) for Svix-based verification.
  - Exposed as resend.webhooks on the client.
  - Includes unit tests covering header mapping and verification flow.

- **Dependencies**
  - Added svix@1.76.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "resend",
-  "version": "6.1.3",
+  "version": "6.2.0",
   "description": "Node.js library for the Resend API",
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",


gabrielmfern:
I think anyone has access to just cherry-pick commits into `main`, feel free to do it, then another PR to bump, and then we can rebase `canary` after

Issue #587, opened 2025-08-21
Bundler error in Cloudflare Workers: Cannot resolve "@react-email/render" when using resend.emails.send with html only (React not used)


nexckycort:
### Describe the Bug

I'm experiencing a bundling error when deploying to Cloudflare Workers using Wrangler v4.32.0. Even though I’m only sending HTML emails via the `html` prop (without any React component), the build fails with:

```
ERROR: Could not resolve "@react-email/render"
```

This happens despite using the latest `resend` version (v6.0.1). If I downgrade `resend` to an older version (4.8.0), the error disappears.

### Link to the code that reproduces this issue

https://github.com/resend

### To Reproduce

**Steps to Reproduce**

1. Install latest `resend` (e.g., `resend@6.0.1`).
2. In a Cloudflare Worker, call:

   ```js
   await resend.emails.send({
     from: "noreply@...com",
     to: email,
     subject: `...`,
     html: "<div>Some HTML</div>",
   });
   ```
3. Run `wrangler deploy` — even without using `react` at all — results in:

   ```
   ERROR: Could not resolve "@react-email/render"
   ```

### Expected Behavior

The Worker should bundle and deploy successfully when using plain HTML only.

jayanratna:
I believe installing `@react-email/render` will fix this issue for you. `npm install @react-email/render` :)

gabrielmfern:
Couldn't reproduce with our [Cloudflare Workers example](https://github.com/resend/resend-cloudflare-workers-example), even after updating to wrangler@4.32.0 and removing `@react-email` completely.

Can you please share a reproduction for this?

gabrielmfern:
Was able to reproduce this when bundling with esbuild, seems like it happens because the bundler tries to bundle in `@react-email/render`. Workaround is to just set `@react-email/render` as external

astanciu:
Having the same issue. @gabrielmfern where did you set it as external?

gabrielmfern:
@astanciu I'm afraid there's no external for the `wrangler` CLI, this only works when doing bundling yourself, or using something more customizable. The simplest thing you can do, *for Cloudflare Workers*, is installing `@react-email/render` so that the error goes away.

chakravarthi-bb:
[...]

PR #699, opened 2025-10-21
feat: bundler (esbuild, nextjs) integration tests

gabrielmfern:
This adds in a new directory `integrations` with a test file `integrations.spec.ts` that runs over the two integrations `integrations/nextjs` and `integrations/esbuild` respectively to make sure we don't introduce regressions that cause #625 or #587 respectively.

To actually get this to work reliably, I had to make sure that `@react-email/render` was not accessible for `esbuild` or `turbopack` builds. To do that, because of [standard module resolution](https://nodejs.org/api/modules.html#all-together) traversing parent `node_modules` directories, the test moves the integration's directory into the developer's `tmp` directory.

    

---
## Summary by cubic
Adds integration tests for Next.js and esbuild to prevent bundler regressions when using the SDK. Tests build minimal apps in an isolated temp directory to avoid parent node_modules leaking (e.g., @react-email/render).

- **New Features**
  - Added integrations folder with minimal Next.js app and esbuild project using Resend.
  - Vitest builds each app from a tmp copy; package.json is rewritten to depend on the local SDK path.
  - New scripts: test:integrations and integration:nextjs; tsconfig now includes integrations.

<sup>Written for commit 6db5bea5a1bed37e83d61814f1a330e100bb3b50. Summary will update automatically on new commits.</sup>

### DIFF
--- a/integrations/integrations.spec.ts
+++ b/integrations/integrations.spec.ts
@@ -24,7 +24,10 @@ describe('integrations', () => {
    */
   async function prepareTemporaryIntegrationCopy(integrationPath: string) {
     const temporaryIntegrationPath = await fs.promises.mkdtemp(
-      `resend-node-integration-${path.basename(integrationPath)}`,
+      path.join(
+        os.tmpdir(),
+        `resend-node-integration-${path.basename(integrationPath)}`,
+      ),
     );
     await fs.promises.cp(
       path.resolve(__dirname, integrationPath),


gabrielmfern:
@SocketSecurity ignore-all

integrations/nextjs/app/route.js
```diff
@@ -0,0 +1,7 @@
+import { Resend } from 'resend';
+
+export function GET() {
```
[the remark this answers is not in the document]
gabrielmfern: Can't, this is Next.js function naming that we can't override.

integrations/integrations.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Create an extra temporary copy of the given integration so that there's no `@react-email/render` module resolution from the resend-node's devDependencies
+   *
+   * Also modifies the package.json to point to the SDK with an absolute path.
+   */
+  async function prepareTemporaryIntegrationCopy(integrationPath: string) {
+    const temporaryIntegrationDir = await fs.promises.mkdtempDisposable(
+      `resend-node-integration-${path.basename(integrationPath)}`,
+    );
```
lucasfcosta: Quick note: `fs.promises.mkdtempDisposable` apparently is only available on Node >= v24 but our package.json requires node: ">=20" so this might fail locally for some people using that version.

Nbd, I think we can leave as is, just noting.

lucasfcosta (approved):

integrations/integrations.spec.ts
gabrielmfern: good catch, going to use the older `mkdtemp`

PR #713, opened 2025-10-24
feat: return response headers on all responses

gabrielmfern:
Most of this is updating tests, the actual main changes are to `resend.ts` and `interfaces.ts`

---
## Summary by cubic
Return HTTP response headers on all SDK responses to improve debugging and integration visibility. Adds a headers field to response objects and unifies types via a new Response<T> generic.

- **New Features**
  - Every response now includes headers: Record<string, string> | null (e.g., authorization, content-type).

- **Refactors**
  - Introduced Response<T> type and replaced per-endpoint unions with this shared type.
  - Updated fetchRequest to attach headers on success and error responses.
  - Revised all interfaces and tests to use the new response shape.

### DIFF
--- a/src/emails/receiving/interfaces/get-receiving-email.interface.ts
+++ b/src/emails/receiving/interfaces/get-receiving-email.interface.ts
@@ -24,4 +24,5 @@ export interface GetReceivingEmailResponseSuccess {
   }>;
 }
 
-export type GetReceivingEmailResponse = Response<GetReceivingEmailResponseSuccess>;
+export type GetReceivingEmailResponse =
+  Response<GetReceivingEmailResponseSuccess>;


src/broadcasts/interfaces/create-broadcast-options.interface.ts
joaopcm, line 31: This link is broken

joaopcm, line 44: This link is broken

joaopcm, line 74: Think link is broken

src/broadcasts/interfaces/update-broadcast.interface.ts
joaopcm, line 12: Should we add the `@deprecated` warning to the create method as well?

src/common/utils/parse-template-to-api-options.ts
joaopcm, line 15: This change is invalid. We don't support `boolean`, `object`, and `list` anymore. Maybe we need to rebase this brach?

joaopcm: Same thing for `fallback_value`, we shouldn't be adding `Record<string, unknown>`, `TemplateVariableListFallbackType` in this case

src/emails/interfaces/create-email-options.interface.ts
joaopcm, line 31: same here, it doesn't seem to be up to date

joaopcm: boolean variables are no longer supported (for now)

PR #717, opened 2025-10-24
refactor: reorganize attachments methods to match final API design

vieiralucas:
- Moved resend.attachments.sending → resend.emails.attachments
- Moved resend.attachments.receiving → resend.emails.receiving.attachments
- Removed top-level resend.attachments module
- Updated API paths from /emails/sending/ to /emails/

Depends on https://github.com/resend/resend-api/pull/2291

---
## Summary by cubic
Reorganized attachments under Emails to match the final API design and align with PRODUCT-791. Endpoints now use /emails/, and the top-level attachments module is removed.

- **Refactors**
  - resend.attachments.sending → resend.emails.attachments
  - resend.attachments.receiving → resend.emails.receiving.attachments
  - Removed top-level resend.attachments
  - Updated endpoints from /emails/sending/... to /emails/...

- **Migration**
  - Replace calls: resend.attachments.sending.* → resend.emails.attachments.*
  - Replace calls: resend.attachments.receiving.* → resend.emails.receiving.attachments.*
  - Stop importing attachments interfaces from the top-level index.
  - Requires resend-api PR #2291 (for /emails/... endpoints).


src/emails/receiving/attachments/interfaces/get-attachment.interface.ts
```diff
@@ -0,0 +1,21 @@
+import type { ErrorResponse } from '../../../../interfaces';
+import type { Attachment } from './attachment';
+
+export interface GetAttachmentOptions {
+  emailId: string;
+  id: string;
+}
+
+export type GetAttachmentResponseSuccess = {
+  object: 'attachment';
```
bukinoshita: This need to be changed
```suggestion
  object: 'email';
```

vieiralucas: Why? Isn't it an `attachment`? Am I missing something?

bukinoshita: The entity is email, not attachment

vieiralucas: The api does return `attachment` in this field. So'll just merge as it is, if you mean that we should also change the API let me know and we can talk about this.

PR #839, opened 2026-02-07
feat: add optional baseUrl and userAgent to Resend constructor

xiaoyu2er:
## Summary
Add optional `options` to the Resend constructor so callers can override `baseUrl` and `userAgent` while keeping the existing default/env behavior.

## Changes
- **ResendOptions**: New optional interface with `baseUrl?: string` and `userAgent?: string`.
- **Constructor**: `new Resend(key?, options?)` — when `options.baseUrl` / `options.userAgent` are provided, they override env and defaults.
- **Instance properties**: `resend.baseUrl` and `resend.userAgent` (readonly) used for requests and headers.
- **Defaults**: Existing logic preserved via `getDefaultBaseUrl()` and `getDefaultUserAgent()` (env `RESEND_BASE_URL` / `RESEND_USER_AGENT` then fallbacks).
- **Exports**: `ResendOptions` exported from the package entry.

## Usage
```ts
// Default (env or https://api.resend.com)
const resend = new Resend('re_xxx');

// Custom baseUrl and/or userAgent
const resend = new Resend('re_xxx', {
  baseUrl: 'https://eu.api.resend.com',
  userAgent: 'my-app/1.0',
});
```

## Tests
- `src/resend.spec.ts`: 10 tests for default values, options override, env override, and that requests use custom baseUrl and User-Agent.

---
## Summary by cubic
Adds optional per-instance `baseUrl` and `userAgent` to `Resend`. Options override env vars and defaults, and are used in request URLs and headers.

- **New Features**
  - Constructor: `new Resend(key?, { baseUrl?, userAgent? })`
  - Options override `RESEND_BASE_URL` and `RESEND_USER_AGENT`; else fall back to env/defaults
  - Readonly properties: `resend.baseUrl` and `resend.userAgent`
  - Export `ResendOptions` from the package entry

<sup>Written for commit b40556b889ec81c4d173287f4e1d2a6d224d602a. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/resend-node/pull/839?utm_source=github">Review in cubic</a></sup>


xiaoyu2er:
could you take a look @cassiozen @gabrielmfern 🙏

gabrielmfern:
What values would be passing in for `baseUrl` though? That `https://eu.api.resend.com` url doesn't really exist, and I don't think there's any other popular API that is Resend compatible, right?

xiaoyu2er:
@gabrielmfern

We added the options parameter so you can configure the client in code instead of having to set `RESEND_BASE_URL` / `RESEND_USER_AGENT` in the environment. That gives:
- Explicit configuration — All config lives where the client is created (`new Resend(apiKey, { baseUrl, userAgent })`), so it’s obvious what each instance uses. No need to remember or document which env vars must be set.
- Easier testing — Tests can pass a mock base URL or custom user agent per test or per suite, without mutating process.env or cleaning it up. No env setup/teardown in tests.
- Multiple clients — You can use different base URLs or user agents in the same process (e.g. prod vs staging, or different regions later) by creating multiple Resend instances with different options. With env-only config, you’d have to change env at runtime.
- Library / embedded use — When Resend is used inside another app or library, the caller can pass options directly instead of requiring the host app to set env vars. The SDK doesn’t depend on “ambient” env.
- Env still supported — If you prefer env (e.g. in serverless or Docker), `RESEND_BASE_URL` / `RESEND_USER_AGENT` still apply when you don’t pass options; explicit options override env, so both styles are supported.

So the benefit is: you can configure the client explicitly at the call site instead of relying on environment variables, which makes behavior clearer, testing simpler, and multi-instance / library use straightforward.

xiaoyu2er:
@gabrielmfern hi, i wonder could you take a look 🙏 look forward to your reply 🙇

gabrielmfern (approved):

PR #853, opened 2026-02-20
fix: prevent payload mutation and parallelize batch rendering

Ray0907:
- **Fix payload mutation** in `emails.create()`, `templates.create()`, and
  `batch.create()` — these methods currently modify the caller's input object
  (`payload.html = ...`, `email.react = undefined`), which can cause bugs when reusing the
  same options object across multiple calls
 - **Parallelize batch rendering** — replace sequential `for...of` + `await` with
  `Promise.all`, so React template rendering for batch emails runs concurrently instead of
  one-by-one

---
## Summary by cubic
Prevents payload mutation across email/template/batch APIs and runs batch React rendering in parallel to speed up sends. Removes side effects when reusing options objects and lowers batch latency.

- **Bug Fixes**
  - Stop mutating caller input (payload.html, email.react) in emails.create(), templates.create(), and batch.create(); use immutable copies.

- **Performance**
  - Parallelize batch React rendering in batch.create() with Promise.all instead of sequential awaits.

<sup>Written for commit 24647f5d24593776407cf49cacba874a12c48126. Summary will update on new commits.</sup>

### DIFF
--- a/src/batch/batch.ts
+++ b/src/batch/batch.ts
@@ -1,4 +1,3 @@
-import type { EmailApiOptions } from '../common/interfaces/email-api-options.interface';
 import { parseEmailToApiOptions } from '../common/utils/parse-email-to-api-options';
 import { render } from '../render';
 import type { Resend } from '../resend';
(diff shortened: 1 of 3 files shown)


felipefreitag:
Thank you, superseded by #862

PR #861, opened 2026-03-02
fix: typo and align remove-contact response with API

Shubhdeep12:
- contacts.create(): fix parameter name typo.
- contacts.remove() response: added `id` to match API; keep `contact` as deprecated alias. we can start using `id` in the new code and `contact` we can remove in the next major.

reference: https://github.com/resend/resend-openapi/blob/main/resend.yaml

---
## Summary by cubic
Aligns remove-contact response with the API by returning id and keeping contact as a deprecated alias for backward compatibility. Also fixes a parameter typo in contacts.create and updates topics tests to match API field names.

- **Bug Fixes**
  - contacts.remove(): return id; keep contact as a deprecated alias; normalize responses for back-compat.
  - contacts.create(): fix payload parameter name typo.
  - Tests updated to expect default_subscription (snake_case) in topics.

- **Migration**
  - Read id from remove-contact responses. contact is deprecated and will be removed in the next major.

<sup>Written for commit d818babf7f14b5ee4affeabd020a106baae306f3. Summary will update on new commits.</sup>

### DIFF
--- a/src/contacts/contacts.spec.ts
+++ b/src/contacts/contacts.spec.ts
@@ -876,11 +876,12 @@ describe('Contacts', () => {
   describe('remove', () => {
     it('removes a contact by id', async () => {
       const response: RemoveContactsResponseSuccess = {
-        contact: '3d4a472d-bc6d-4dd2-aa9d-d3d50ce87223',
+        id: '3d4a472d-bc6d-4dd2-aa9d-d3d50ce87223',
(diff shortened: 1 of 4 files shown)


dielduarte:
hey thanks for the contribution, but we feel like contact should be the one used here. as id is also used elsewhere, causing  unnecessary confusion

PR #862, opened 2026-03-02
fix: avoid mutating payloads in emails, broadcasts, and templates

Shubhdeep12:
- used cloned bodies / local html variables instead of writing payload.html when react is provided.
- added a test

---
## Summary by cubic
Stop mutating request payloads in emails, broadcasts, and templates when rendering React to HTML. This prevents side effects and keeps input objects unchanged.

- **Bug Fixes**
  - Use local html variables/body copies instead of writing to payload.html in emails.create, broadcasts.create/update, and templates.create.
  - Added tests to verify payload immutability for emails, broadcasts, and templates. No API changes.

<sup>Written for commit c6dcfb439099486b816dd4e6b6b9bd5f7c11bd01. Summary will update on new commits.</sup>

### DIFF
--- a/src/templates/templates.spec.ts
+++ b/src/templates/templates.spec.ts
@@ -247,6 +247,34 @@ describe('Templates', () => {
       expect(mockRenderAsync).toHaveBeenCalledWith(mockReactComponent);
     });
 
+    it('does not mutate payload when using React component', async () => {
+      const mockReactComponent = {
+        type: 'div',
+        props: { children: 'Welcome!' },
+      } as React.ReactElement;
+
+      mockRenderAsync.mockResolvedValueOnce('<div>Welcome!</div>');
+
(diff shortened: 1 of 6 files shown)


src/templates/templates.spec.ts
```diff
@@ -247,6 +247,34 @@ describe('Templates', () => {
       expect(mockRenderAsync).toHaveBeenCalledWith(mockReactComponent);
     });
 
+    it('does not mutate payload when using React component', async () => {
```
gabrielmfern, line 250: can you also add tests to email.ts and @broadcasts.ts for this?

Shubhdeep12: added. thanks for review.

gabrielmfern (approved):
Thank you!

PR #863, opened 2026-03-02
fix: to support @react-email/render exports across versions in templates

Shubhdeep12:
- `@react-email/render` dropped renderAsync in v2.0.0.

---
## Summary by cubic
Use a shared `render` helper to render React components in Templates, aligning with `@react-email/render`’s `render` API and removing dynamic imports. This simplifies Templates (and Emails) and preserves a clear error when `@react-email/render` is missing.

- **Bug Fixes**
  - Templates now call the shared `render` directly; dynamic import and cached renderer removed.
  - Clear error when `@react-email/render` isn’t installed.
  - Tests updated to mock `render` (removed `renderAsync` usage).
  - Minor cleanup in Emails: removed unnecessary React type cast.

<sup>Written for commit a19455545da2711c99271cb5b57703bae8021b0a. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/resend-node/pull/863?utm_source=github">Review in cubic</a></sup>


src/templates/templates.ts
```diff
@@ -54,12 +54,23 @@ export class Templates {
     if (payload.react) {
       if (!this.renderAsync) {
         try {
-          const { renderAsync } = await import('@react-email/render');
-          this.renderAsync = renderAsync;
-        } catch {
-          throw new Error(
-            'Failed to render React component. Make sure to install `@react-email/render`',
-          );
+          const mod = await import('@react-email/render');
```
gabrielmfern: this should use

https://github.com/resend/resend-node/blob/d16e477ba125a153876cd9393a85a4c6fe53b047/src/render.ts#L1-L6

gabrielmfern (changes requested):

src/templates/templates.ts
Shubhdeep12: yeah, i thought earlier, since it's a peer dep, we wanted to give support for renderAsync.

thanks for the review, made the changes.

gabrielmfern: that's valid, but it should be done in that render function and reused

src/templates/templates.ts
```diff
@@ -46,26 +46,13 @@ export class Templates {
...
-          throw new Error(
-            'Failed to render React component. Make sure to install `@react-email/render`',
-          );
-        }
-      }
-
-      payload.html = await this.renderAsync(
-        payload.react as React.ReactElement,
-      );
+      payload.html = await render(payload.react as React.ReactElement);
```
gabrielmfern: I don't think this cast is required, is it?

Shubhdeep12: yeah we can remove it, especially as we are narrowing the scope here with ReactElement. 
lemme remove it and from other places as well.

src/templates/templates.ts
Shubhdeep12: i removed it in the last commit, lemme add the support of `renderAsync` back
 and to `render.ts` that the right place

src/render.ts
```diff
@@ -1,12 +1,24 @@
...
     throw new Error(
       'Failed to render React component. Make sure to install `@react-email/render` or `@react-email/components`.',
     );
   }
 
-  return render(node);
+  const renderFn = mod.render ?? mod.renderAsync;
+  if (typeof renderFn !== 'function') {
+    throw new Error(
+      'Failed to render React component. Make sure to install `@react-email/render` (or an older version with renderAsync).',
```
gabrielmfern:
```suggestion
      'Could not find the React Email render function, this is most likely a bug, please file an issue.',
```

felipefreitag (approved):

gabrielmfern (approved):

PR #963, opened 2026-05-18
feat: add `ContactImports` class for managing contact imports

vcapretz:
## Summary by cubic
Adds `ContactImports` under `resend.contacts.imports` to create, list, and get contact import jobs with multipart CSV uploads, pagination, and status filters. Updates the client to handle `FormData` correctly and exports the new interfaces.

- **New Features**
  - Added `contacts.imports` (`create`, `list`, `get`) via `/contacts/imports`; `create` supports multipart uploads (`file`) with `columnMap`, `onConflict`, `onError`, `segments`, and `topics`; `list` supports pagination and `status` (`queued`, `in_progress`, `completed`, `failed`).
  - Exposed on `Contacts` as `imports`; exported interfaces via `index`.

- **Bug Fixes**
  - `Resend.post` now detects `FormData`, removes `Content-Type`, and sends the form body as-is to support multipart uploads.

<sup>Written for commit c778b2fbdc1e57f9d712a74a331f26cf22881073. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/resend-node/pull/963?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/contacts/imports/contact-imports.ts
+++ b/src/contacts/imports/contact-imports.ts
@@ -1,0 +1,28 @@
+import { buildPaginationQuery } from '../../common/utils/build-pagination-query';
+import type { Resend } from '../../resend';
+import type {
+  ListContactImportsOptions,
+  ListContactImportsResponse,
+  ListContactImportsResponseSuccess,
+} from './interfaces/list-contact-imports.interface';
+
+export class ContactImports {
+  constructor(private readonly resend: Resend) {}
+
+  async list(
+    options: ListContactImportsOptions = {},
+  ): Promise<ListContactImportsResponse> {
+    const searchParams = new URLSearchParams(buildPaginationQuery(options));
+
+    if (options.status !== undefined) {
+      searchParams.set('status', options.status);
+    }
+
+    const queryString = searchParams.toString();
+    const url = queryString
+      ? `/contacts/imports?${queryString}`
+      : '/contacts/imports';
+
+    return this.resend.get<ListContactImportsResponseSuccess>(url);
+  }
+}
--- a/src/contacts/imports/contact-imports.ts
+++ b/src/contacts/imports/contact-imports.ts
@@ -1,6 +1,17 @@
 import { buildPaginationQuery } from '../../common/utils/build-pagination-query';
 import type { Resend } from '../../resend';
 import type {
+  ContactImportColumnMap,
+  CreateContactImportOptions,
+  CreateContactImportRequestOptions,
+  CreateContactImportResponse,
+  CreateContactImportResponseSuccess,
+} from './interfaces/create-contact-import.interface';
+import type {
+  GetContactImportResponse,
+  GetContactImportResponseSuccess,
+} from './interfaces/get-contact-import.interface';
+import type {
   ListContactImportsOptions,
   ListContactImportsResponse,
   ListContactImportsResponseSuccess,
@@ -8,6 +19,19 @@ import type {
 
 export class ContactImports {
   constructor(private readonly resend: Resend) {}
+
+  async create(
+    payload: CreateContactImportOptions,
+    options: CreateContactImportRequestOptions = {},
+  ): Promise<CreateContactImportResponse> {
+    const formData = this.buildCreateFormData(payload);
+
+    return this.resend.post<CreateContactImportResponseSuccess>(
+      '/contacts/imports',
+      formData,
+      options,
+    );
+  }
 
   async list(
     options: ListContactImportsOptions = {},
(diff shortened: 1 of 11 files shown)


lucasfcosta (approved):
Nice!

vcapretz:
example script to create a new import:


```typescript
#!/usr/bin/env node

import { openAsBlob } from 'node:fs';
import { access } from 'node:fs/promises';
import path from 'node:path';
import { Resend } from '../dist/index.mjs';

const apiKey = process.env.RESEND_API_KEY;

if (!apiKey) {
  console.error('Missing RESEND_API_KEY environment variable.');
  process.exit(1);
}

const csvPath = path.resolve(process.cwd(), 'contacts.csv');

try {
  await access(csvPath);
} catch {
  console.error(`CSV file not found: ${csvPath}`);
  process.exit(1);
}

const file = await openAsBlob(csvPath, { type: 'text/csv' });
const resend = new Resend(apiKey);

const result = await resend.contacts.imports.create({
  file,
  fileName: path.basename(csvPath),
  columnMap: { email: 'Email', firstName: 'Name', lastName: 'Last Name' },
  onConflict: 'skip',
  onError: 'continue',
});

console.log(JSON.stringify(result, null, 2));

if (result.error) {
  process.exitCode = 1;
}
```

src/contacts/imports/interfaces/create-contact-import.interface.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface ContactImportTopic {
+  id: string;
+  subscription: 'opt_in' | 'opt_out';
+}
+
+export interface CreateContactImportOptions {
+  file: Blob;
+  fileName?: string;
```
lucasfcosta: Why do we need filename? It's not used at all by the back-end. We only need the actual file content.

src/contacts/imports/contact-imports.ts
```diff
@@ -0,0 +1,114 @@
...
+    return this.resend.get<GetContactImportResponseSuccess>(
+      `/contacts/imports/${id}`,
+    );
+  }
+
+  private buildCreateFormData(payload: CreateContactImportOptions): FormData {
+    const formData = new FormData();
+
+    if (payload.fileName !== undefined) {
+      formData.append('file', payload.file, payload.fileName);
```
lucasfcosta: Do we even need this `fileName` check? See my comment above on `fileName` not being necessary.

src/contacts/imports/contact-imports.ts
```diff
@@ -0,0 +1,114 @@
...
+    );
+    this.appendStringField(formData, 'on_conflict', payload.onConflict);
+    this.appendStringField(formData, 'on_error', payload.onError);
+    this.appendJsonField(formData, 'segments', payload.segments);
+    this.appendJsonField(formData, 'topics', payload.topics);
+
+    return formData;
+  }
+
+  private buildColumnMap(columnMap?: ContactImportColumnMap) {
```
lucasfcosta: NIT: I usually don't like optional params. I prefer passing explicit `null`s in so that we have defensive types instead of forgetting to pass something and relying on default behavior. Won't block because of this, I'll leave it up to you.

src/contacts/imports/contact-imports.ts
```diff
@@ -0,0 +1,114 @@
...
+  private appendStringField(
+    formData: FormData,
+    name: string,
+    value: string | undefined,
+  ): void {
+    if (value !== undefined) {
+      formData.append(name, value);
+    }
+  }
+}
```
lucasfcosta: Can't we use just one function and run `typeof` to decide whether to stringify?

src/resend.ts
```diff
@@ -166,6 +166,13 @@ export class Resend {
     options: PostOptions & IdempotentRequest = {},
   ) {
     const headers = new Headers(this.headers);
+    const isFormData =
+      typeof FormData !== 'undefined' && entity instanceof FormData;
```
lucasfcosta, line 170: `FormData` is not supported on Node 16 and below, which I think it's fine as it's out of LTS and was released in 2021 and LTS ended in 2023, but just noting.

lucasfcosta (approved):
No blockers, but I think the `fileName` comment is worth looking at. Same for other cleanups.

src/resend.ts
vcapretz: oh yeah, at this point i wouldn't say we support Node versions older than 20 (maybe even 22)

lucasfcosta (approved):

PR #970, opened 2026-05-25
fix: replace svix with standardwebhooks to reduce install size (#969)

### LINKED ISSUE #969: Package size much bigger
### Discussed in https://github.com/resend/resend-node/discussions/838

<div type='discussions-op-text'>

<sup>Originally posted by **letkeeko** February  7, 2026</sup>
Hello guys, what happened to the package size?

Last time I used resend (v6.1.2) about three months ago for a toy project, it was much smaller.
Just come back to restart the project and installed resend (v6.9.1), and was surprised by the difference in package size...

Screenshot 2026-02-07 at 4 28 46 pm
Screenshot 2026-02-07 at 4 28 14 pm
</div>

dielduarte:
## Problem

[#969](https://github.com/resend/resend-node/issues/969) reports that the installed package size grew dramatically. Measured on a clean install:

| Version | Total `node_modules` |
|---|---|
| `resend@6.1.2` | 136 KB |
| `resend@6.12.3` (before this PR) | **5.0 MB** (~37×) |

The overwhelming majority is **`svix` (4.2 MB, 84% of the install)**, which is pulled in **only** for webhook signature verification in `src/webhooks/webhooks.ts`.

## Why svix is so heavy (and why tree-shaking doesn't help)

- The `Webhook` class resend uses is a 28-line wrapper that just maps `svix-*` headers to `webhook-*` and **delegates to [`standardwebhooks`](https://www.npmjs.com/package/standardwebhooks) (~13 KB)**.
- The rest of svix is the full **Svix platform API client** — 216 OpenAPI-generated model files (`dist/models` alone is 2.7 MB) plus 23 API resource clients — none of which resend uses.
- svix is `"type": "commonjs"` with no `module`/`exports`/`sideEffects`, and its barrel `index.js` eagerly `require()`s the entire API client, so it **cannot be tree-shaken**. (And tree-shaking never affects on-disk install size anyway — `npm install` always downloads the whole tarball.)

## Change

Depend on **`standardwebhooks` directly** instead of `svix`:

- `package.json`: `svix@1.92.2` → `standardwebhooks@1.0.0`
- `src/webhooks/webhooks.ts`: import `Webhook` from `standardwebhooks`; pass `webhook-id` / `webhook-timestamp` / `webhook-signature` (the standard header names svix mapped to internally).

`standardwebhooks` is the same pure-JS HMAC-SHA256 implementation svix delegates to (keeps Cloudflare Workers / non-Node support) and produces **identical signatures**.

### No public API change
`Webhooks.verify(...)` and `VerifyWebhookOptions` (`headers.{ id, timestamp, signature }`) are unchanged.

## Result

Install size **5.0 MB → 788 KB (~84% smaller)**, verified via `npm pack` + clean install:

```
node_modules/postal-mime     352K
node_modules/resend          244K
node_modules/@stablelib       96K
node_modules/standardwebhooks 44K
node_modules/fast-sha256      44K
-------------------------------------
total                        788K
```

## Tests

- Updated `webhooks.spec.ts` to mock `standardwebhooks` and assert the `webhook-*` header names.
- Added `webhooks-verify.spec.ts`: a **real, non-mocked round-trip test** that signs a payload with the genuine library and asserts `verify()` returns the parsed event (and rejects a bad signature). The prior tests fully mocked verification, so the real crypto path was untested.
- `pnpm test` ✅ 358 passed · `pnpm run build` ✅ · `pnpm run lint` ✅

Closes #969

---
## Summary by cubic
Replace `svix` with `standardwebhooks` for webhook signature verification to shrink install size from ~5.0 MB to ~788 KB (~84%) with no behavior or API changes. Addresses DEV-848 and #969; verification now uses `webhook-*` headers internally, and tests are consolidated into `webhooks.spec.ts` using a real sign+verify round trip plus a bad‑signature case (no mocks).

<sup>Written for commit 014ee22afa21e9d51bb715a7b1f41cb9bd4af7ba. Summary will update on new commits. <a href="https://cubic.dev/pr/resend/resend-node/pull/970?utm_source=github">Review in cubic</a></sup>


dielduarte:
> [Open in StackBlitz](https://pkg.pr.new/template/c8dfb936-6cba-4c6e-aa3d-2dc5a4a7100b)
> 
> ```
> npm i https://pkg.pr.new/resend@970
> ```
> 
> _commit: [`254b263`](https://github.com/resend/resend-node/runs/77724241418)_

verified locally using this version, and it works as expected! no public API changes and no regressions. cc @gabrielmfern

gabrielmfern (approved):

src/webhooks/webhooks-verify.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { Webhook } from 'standardwebhooks';
+import { describe, expect, it } from 'vitest';
+import { Resend } from '../resend';
+
+// Uses the REAL `standardwebhooks` implementation (no module mock) to prove the
+// end-to-end signature verification path works after migrating off `svix`.
+describe('Webhooks.verify (real signature round-trip)', () => {
```
gabrielmfern: why not have this test in webhooks.spec?

dielduarte: great catch, fixed here; [014ee22](https://github.com/resend/resend-node/pull/970/commits/014ee22afa21e9d51bb715a7b1f41cb9bd4af7ba)

src/webhooks/webhooks.ts
```diff
@@ -91,9 +91,9 @@ export class Webhooks {
   verify(payload: VerifyWebhookOptions): WebhookEventPayload {
     const webhook = new Webhook(payload.webhookSecret);
     return webhook.verify(payload.payload, {
-      'svix-id': payload.headers.id,
-      'svix-timestamp': payload.headers.timestamp,
-      'svix-signature': payload.headers.signature,
+      'webhook-id': payload.headers.id,
```
gabrielmfern, line 94: this property name is intentional right? these are just properties related to svix?

dielduarte: yes, because internally it used to map svix -> webhook, now that we use the standard directly it is webhook

lucasfcosta (approved):
LGTM! Good catch using the same library Svix does :)

PR #1047, opened 2026-08-06
fix: replace deprecated node10 moduleResolution

ffakira:
## Problem

`pnpm typecheck` fails on `canary`. TypeScript 6.0 escalates `moduleResolution: "node"` (node10) from a warning to error TS5107:

```
tsconfig.json(9,25): error TS5107: Option 'moduleResolution=node10' is deprecated
and will stop functioning in TypeScript 7.0.
```

Because this is a config-level error, `tsc` aborts before checking any file, which was masking 11 pre-existing type errors.

## Change

Switch to `moduleResolution: "bundler"`. This requires `module` to be `preserve`/`es2015+`, so both move together. `preserve` is the right pairing here: `noEmit` is set, so `module` only affects typecheck semantics, tsdown performs the actual dual ESM / CJS emit and doesn't consult this option.

The resolution switch introduces **no new errors**. Running the previous node10 config with `ignoreDeprecations: "6.0"` produces a byte-identical error set; the config error was only hiding what was already there. All 11 are fixed:

- `src/resend.ts` — two error literals in `fetchRequest` lacked an `ErrorResponse` annotation, so `name` widened to `string` rather than `RESEND_ERROR_CODE_KEY`. Those failure paths returned values that didn't conform to the public `ErrorResponse` type. A third literal in the same function was already annotated; this makes all three consistent.
- `CreateBatchEmailOptions` — see below.
- `topics.spec.ts` — three response fixtures used `defaultSubscription`; the API returns `default_subscription` ([docs](https://resend.com/docs/api-reference/topics/get-topic)). Six sites change, since each fixture also has a matching `toMatchInlineSnapshot`. Worth noting these tests were passing while asserting a shape the API never returns. The `create()` *request* options are correctly camelCase and are unchanged.
- `receiving.spec.ts` — two list mocks omitted the required `received_for`; set to `[]`, matching the eight other mocks in the file. Four sites change, again because of the paired inline snapshots.
- `tsconfig.json` — excludes `e2e/esbuild` and `e2e/nextjs`. These aren't workspace members (`pnpm-workspace.yaml` has no `packages:` key); `e2e.spec.ts` copies them to a temp dir and runs `npm install --install-links`, so `import { Resend } from 'resend'` never resolved from the repo. They're verified by `pnpm test:e2e`. `e2e/e2e.spec.ts` itself stays type-checked.

## This tightens a published type

`CreateBatchEmailOptions` was `Omit<CreateEmailOptions, 'attachments'>`, but the built-in `Omit` doesn't distribute over unions: `Omit<A | B, K>` collapses both branches into one flat object, erasing the mutual exclusivity between the `react`/`html`/`text` variant and the `template` variant.

In practice `resend.batch.send()` accepted payloads mixing `template` with `html`, and payloads with no message body at all. A local `DistributiveOmit` maps each union member individually. A type-level probe confirms the two valid shapes still compile and the two invalid ones are now rejected; previously all four compiled.

No runtime behaviour changes and no test depended on the loose type, but consumers relying on the collapsed union will see a new type error. **Suggest releasing as a minor rather than a patch.**

`DistributiveOmit` is kept file-local rather than placed in `common/interfaces`, since everything there is re-exported from the package entry point and would become public API for a single-use helper. Happy to promote it if preferred.

## Verification

|                  |                                        |
| ---------------- | -------------------------------------- |
| `pnpm typecheck` | 0 errors (was: failed at config level) |
| `pnpm build`     | ✅                                      |
| `pnpm test`      | 385 passed, 4 todo                     |
| `pnpm lint`      | ✅                                      |

Notes on `bundler` vs `nodenext`, CI coverage, and one out-of-scope finding in a follow-up comment.


gabrielmfern (approved):

gabrielmfern:
tysm!

gabrielmfern:
also added a tsc on CI for this to not happen again

PR #1072, opened 2026-08-16
feat: request cancellation 

Nithin0620:
## Summary
Adds three optional request controls to the SDK, all **opt-in and backward compatible**
(no behavior change unless you set them):
- **`timeoutMs`** — abort a request attempt after N ms (per attempt, so retries still get a fair shot)
- **`retries`** — retry retryable failures (HTTP 429, 5xx, network errors) with exponential backoff, honoring the `Retry-After` header
- **`signal`** — pass any `AbortSignal` to cancel a request
Set them globally in the constructor or per call:
```ts
const resend = new Resend('re_xxx', { timeoutMs: 10_000, retries: 3 });
// per-call override
await resend.emails.send({ ... }, { timeoutMs: 30_000, retries: 1 });
```

### Why
Currently a request can hang indefinitely and rate limits / transient 5xx fail
immediately with no retry. These options bring the SDK in line with how the other
official Resend SDKs and major API clients behave.

### Notes
- Timeouts and aborts are not retried.
- retries defaults to 0 — retrying a POST (e.g. emails.send) is the caller's
  explicit choice, so idempotency-sensitive flows are unaffected by default.
- No new dependencies, no API surface removed.

### Tests
- 5 new unit tests (src/resend.spec.ts): timeout abort, AbortSignal cancel,
  429 retry with Retry-After, 5xx retry, no retry on 4xx, per-request override.
- pnpm test (394 passed), pnpm typecheck, pnpm lint all green.

### Commits
1. feat: add timeout, retries, and AbortSignal options to requests — the feature core (10 files)
2. fix: forward per-request options in list methods — 21 list methods were silently
   dropping per-request options (get(url) → get(url, options)); this makes
   emails.list({ retries: 2 }) actually work. The bulk of the file count is this
   one-line mechanical change.

---
## Summary by cubic
Adds opt-in request cancellation to the SDK via `AbortSignal` and fixes list methods to forward per-request options. Previously list endpoints ignored per-request options, so a signal passed to `list()` had no effect; now `signal` is forwarded to `fetch` and can cancel requests.

- New `RequestOptions` interface is added to `GetOptions`, `PostOptions`, `PutOptions`, `PatchOptions`, `DeleteOptions`, and `PaginationOptions`.
- List methods now pass `options` to `resend.get()` so per-request signals apply.
- Use `AbortSignal.timeout(ms)` for timeouts; no SDK-level retry logic is included.

<sup>Written for commit 398ae1d2b67799f977a12fa424c783c50779ef7d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/resend/resend-node/pull/1072?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -134,6 +134,27 @@ console.log(`Email ${data.id} with a React template has been sent`);
 >});
 >```
 
+## Request options
+
+`timeoutMs` and `retries` can be set as defaults for all requests, or overridden per call:
+
+```ts
+const resend = new Resend('re_xxx', {
+  timeoutMs: 10_000, // abort each request attempt after 10s
+  retries: 3, // retry 429s, 5xx, and network errors
+});
+
+// override for a single request
+await resend.emails.send({ ... }, { timeoutMs: 30_000, retries: 1 });
+```
+
+When retrying, a `Retry-After` response header is honored; otherwise exponential backoff is applied. Timeouts and aborts are not retried. To cancel a request manually, pass an `AbortSignal`:
+
+```ts
+const controller = new AbortController();
+await resend.emails.send({ ... }, { signal: controller.signal });
+```
+
 ## License
 
 MIT License
--- a/src/api-keys/api-keys.ts
+++ b/src/api-keys/api-keys.ts
@@ -35,7 +35,10 @@ export class ApiKeys {
   async list(options: ListApiKeysOptions = {}): Promise<ListApiKeysResponse> {
     const url = buildPaginationUrl('/api-keys', options);
 
-    const data = await this.resend.get<ListApiKeysResponseSuccess>(url);
+    const data = await this.resend.get<ListApiKeysResponseSuccess>(
+      url,
+      options,
+    );
     return data;
   }
 
--- a/src/automation-runs/automation-runs.ts
+++ b/src/automation-runs/automation-runs.ts
@@ -41,7 +41,10 @@ export class AutomationRuns {
       ? `/automations/${options.automationId}/runs?${qs}`
       : `/automations/${options.automationId}/runs`;
 
-    const data = await this.resend.get<ListAutomationRunsResponseSuccess>(url);
+    const data = await this.resend.get<ListAutomationRunsResponseSuccess>(
+      url,
+      options,
+    );
     return data;
   }
 }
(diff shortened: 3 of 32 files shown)


gabrielmfern:
Thank you so much for this; this looks really interesting!

Three things that come to mind:
- I think we can consider this a breaking change as-is, since it changes how certain requests are handled and would override whatever handling the users would be doing normally, like if they already have their own retry logic. We should be making this opt-in first, then opt-out on a major version
- What's the use case for `timeoutMs`? If there isn't one, I think we'd rather not have it
- There is more than one feature in this pull request: custom timeout, retrying, request canceling. Can you make separate pull requests for each so that it's easier to review?

Nithin0620:
Thanks for the feedback! Makes total sense.

1. **`timeoutMs`:** I agree with dropping it. Supporting `signal` (`AbortSignal`) covers timeouts natively via `AbortSignal.
  timeout(ms)`.
2. **Splitting the PR:** I will update this PR to focus only on request cancellation (`signal`) and remove the other parts.
3. **Retries (Next PR):** I'll open a separate PR for the retry feature and make it opt-in (e.g., `autoRetry: false` by default with a config option to enable it) so it's not a breaking change for existing users.

Nithin0620:
Hi @gabrielmfern 
I've updated this PR based on the feedback:

1. **Scoping down to Request Cancellation (`signal`) only:**
       - Removed all custom retry logic, exponential backoff, and header handling from this PR. I will open a separate PR for retries with an explicit opt-in configuration so it remains non-breaking for existing users.
2. **Dropped `timeoutMs`:**
       - Removed custom `timeoutMs` in favor of standard `AbortSignal.timeout(ms)`, which works natively with `signal`.
3. **Docs & Tests updated:**
       - Updated the README examples to show `signal` usage and native timeouts via `AbortSignal.timeout(...)`.
       - Updated unit tests to cover `AbortSignal` request cancellation.

    All tests, type checks, and linting pass cleanly. Ready for another look when you have a moment!

